PR #16, opened 2021-08-31
Set generation

GuillaumeGen:
Bound sets should not be generated if they are useless.


VictorCMiraldo (approved):

GuillaumeGen:
Before this, `consIsRecursive` was both inspecting the arguments and the return type. But since the return type is always the one we are looking for, all constructors were considered as "recursive", which is not the expected behaviour.
If no constructor is "recursive", then no bounded set is generated (this part of the code was already here, but useless, since `consIsRecursive` was equivalent to `const True` ).

PR #20, opened 2021-09-01
Adding an option --ty-spz ALL and adding the specializer to the tests

GuillaumeGen:


src/Pirouette/Specializer.hs
```diff
@@ -6,9 +6,10 @@ module Pirouette.Specializer
...
 import Pirouette.Specializer.Basics
 
-allSpz :: String -> TypeSpecializer
-allSpz "Bool"   = boolSpz
-allSpz "List"   = listSpz
-allSpz "Unit"   = unitSpz
-allSpz "Tuple2" = tuple2Spz
-allSpz "Maybe"  = maybeSpz
\ No newline at end of file
+allSpz :: String -> Maybe TypeSpecializer
```
VictorCMiraldo, line 9: Thank you! I was about to send a similar patch!

src/Pirouette/Term/ToTla.hs
```diff
@@ -34,6 +34,7 @@ import           Control.Monad.Reader
 import           Control.Monad.Except
 import           Control.Monad.State
 import           Control.Arrow (first, second, (***))
+import           Control.Exception ( catch )
```
VictorCMiraldo, line 37: Is this really needed? you added the import but didn't use the imported function in the change to `mkTLATySpecializer`

```suggestion
```

VictorCMiraldo (approved):
ty

src/Pirouette/Term/ToTla.hs
GuillaumeGen: Yes, you are right, I added it and then changed my mind and did not use it. It should be suppressed.

PR #24, opened 2021-09-02
Refactor and reorganize the handling of even-odd mutual recursion

VictorCMiraldo:
Currently, pirouette can only handle even-odd mutual recursion. The previous sorting and organization of term dependencies grew a little too complex from many organic modifications. I decided it was time to refactor and re-organize some of that.

The previous `processDecls` function has been split in multiple parts:

1. `sortDeps` is now at `Pirtla.Monad` ([here](https://github.com/tweag/pirouette/pull/24/files#diff-f916a708ed6174a1233f4bd8eda2056f01c9fedd5dc9fe735b5aa28bb0909b90R181))
2. resolving the mutual recursion within dependency classes is now in a new module, `Pirtla.Transformations`
3. The term transformations that we want to apply to the definitions required by the transition function will be applied separately from resolving mutual dependencies


VictorCMiraldo:
There's something else off too!

This is the resulting `AddSignature` action from the multisig contract:
```tla
ActionAddSignature(pk1110) ==
  /\ st.cons = "State1"
  /\ LET ds1101 ==
           st.arg0 
         ds1102 ==
           st.arg1
      IN /\ ds1101.cons = "CollectingSignatures"
         /\ LET pmt1107 ==
                  ds1101.arg0 
                pks1108 ==
                  ds1101.arg1
             IN /\ Bool_match(Params_match(param,
                                           LAMBDA sigs957,ds958 : foldr0
                                                                  (LAMBDA a,
                                                                          b : Bool_match
                                                                              (IF pk1110 = a
                                                                               THEN True
                                                                               ELSE False,
                                                                               True,
                                                                               b),
                                                                   False,
                                                                   sigs957)),
                              Bool_match(foldr0(LAMBDA a,b : Bool_match
                                                             (IF a = pk1110
                                                              THEN True
                                                              ELSE False,
                                                              True,
                                                              b),
                                                False,
                                                pks1108),
                                         False,
                                         True),
                              False)
                /\ LET res ==
                         Tuple21(TxConstraints1(Cons(MustBeSignedBy(Unit1), \* <--- HERE! Should be pk1110 not Unit1 !! 
                                                     Nil),
                                                Nil,
                                                Nil),
                                 State1(CollectingSignatures(pmt1107,
                                                             Cons(pk1110,
                                                                  pks1108)),
                                        ds1102))
                    IN st' = res[2] /\ txConstr' = res[1] 
```

GuillaumeGen (approved):

PR #27, opened 2021-09-03
Creating a unit test for cfoldMap and adding it to our test suite

GuillaumeGen:


tests/integration/run-tests.sh
```diff
@@ -234,6 +234,9 @@ run-single-test() {
 
 mecho blue "Building pirouette"
 cabal build
+mecho blue "Run unit tests"
```
VictorCMiraldo, line 237: Maybe do something like:

```
if $onci; then
    cabal test
fi
```

This way we don't run the cabal tests every single time we call `run-tests.sh`

tests/unit/Pirouette/Term/TransformationsSpec.hs
```diff
@@ -102,6 +103,14 @@ spec = do
 
     mapM_ dnfExpected tDNFs
 
+  describe "cfoldmapSpecialize" $ do
+    it "works for cfoldmapSpecialize" $
+      runPrtTest
+        (mapM_ declareDummyFunc ["foldr0", "Bool_match", "True", "False"] >> cfoldmapSpecialize tcfoldBool)
+        `shouldBe` tcfoldBoolSpec
+  --  it "works for cfoldmapSpecialize" $
```
VictorCMiraldo, line 111: Are these comments still needed?

VictorCMiraldo (approved):

Issue #25, opened 2021-09-02
`cfoldmapSpecialize` does not properly lift de Bruijn indices


GuillaumeGen:
Currently, the transition obtained after our transformations is wrong for the MultiSigStateMachine contract.
Indeed, it outputs:
```diff
transition :=
...
  λ (params1098 : Params0) (ds1099 : State0 MSState) (i1100 : Input)
...
      (λ (ds1101 : MSState) (ds1102 : List0 (Tuple20 ByteString
                                                     (List0 (Tuple20 ByteString
                                                                     Integer))))
...
           (λ (pmt1107 : Payment0) (pks1108 : List0 ByteString)
...
                (λ pk1110 : ByteString
...
                         (Bool_match (foldr0 @ByteString
                                 @Bool
                                 (λ (a : ByteString) (b : Bool)
                                  . Bool_match (b/ifThenElse @Bool
!                                         (b/equalsByteString 1#a 0#pk1110)
                                          True
                                          False)
                                      @Bool
                                      True
                                      0#b)
                                 False
                                 1#pks1108)
                             @Bool
                             False
                             True)
                         False)
```
This is caused, by the `cfoldmapSpecialize` function, since if we disable it (by commenting the associated line in the `Main`, the term obtained does not contains any wrong de Bruijn index.

GuillaumeGen:
Solved by #28

PR #28, opened 2021-09-03
Lifting in cfoldmapSpecialize was much simpler than expected

GuillaumeGen:
This address #25 
I expect #27 to fail CI because of this de Bruijn index bug. This should fix it.


VictorCMiraldo (approved):

VictorCMiraldo:
Thanks! Approving and merging!

Issue #12, opened 2021-08-30
Replace `--with-args` and `--pull-destr` with a more flexible mechanism


VictorCMiraldo:
Consider the `MultiSigStateMachine` example. We run pirtla with options `-a param,st,i -p 2`, and we still need to manually alter the generated TLA+ to actually declare `param`. It would be nice if we could improve that and provide something like:

```
-a p:param, v:st, i:inp
```

Indicating that: `p:param` says the first argument should be translated and passed around as a parameter named `param`. 
The second `v:st` says that `st` will be a variable in the skeleton file and finally, `i:inp` says that the the third parameter will be named `inp` and represents the user input. Here is where I'd really like if pirouette automatically pulled the destructor for `inp` to the root of the term.

Then, because we've given a `p:_` type of parameter, pirouette should generated a parameterized `Next` action:

```tlaplus
Next(param) == ....
```

This enables us to easily pass around different parameters and also enables us to define the parameter in the skeleton file without having to [sed into the generated file](https://github.com/tweag/pirouette/blob/main/tests/integration/all.json#L78).

PR #29, opened 2021-09-03
Match on input at the head of transition function

GuillaumeGen:
Whenever the user declares `--prefix transition --with-args st, INPUT, param`, the function translated will be
```
transition' st INPUT param =
  match INPUT#1 with
    C1 x -> transition st#3 (C1 x#0) param#1
    C2 x y -> transition st#4 (C2 x#1 y#0) param#2
```
This transformation of the transition function allows us to get rid of the so-called "destructor normal form" and of the `--pull-destr` option.


VictorCMiraldo:
I like where you're going with this, but I think we might be better off writing it as a transformation in `Pirouette.Term.Transformation`. We could call it `reverseEta` or `reverseCase`, and give it the necessary parameters.

GuillaumeGen:
Are those modifications what you expected?

src/Pirouette/Term/Transformations.hs
```diff
@@ -295,6 +296,92 @@ constrDestrId = pushCtx "constrDestrId" . rewriteM (runMaybeT . go)
...
+
+    -- `blindDest out ty` constructs the function
+    -- \ (i : ty) (f : ty -> out) ->
+    -- match i with
+    --   C1 x0 -> f (C1 x0)
+    --   C2 x0 x1 -> f (C2 x0 x1)
+    blindDest :: PrtType -> PrtTypeDef -> PrtTerm
+    blindDest tyOut (Datatype _ _ dest cons) =
+      R.Lam (R.Ann $ fromString "i") (error "Fictive type of i") $ -- Since the generated term will be applied to two arguments,
+        R.Lam (R.Ann $ fromString "f") (error "Fictive type of f") $ -- those 2 types should never be accessed.
```
VictorCMiraldo, line 341: I understand why you put an error here and I respect it if you rather merge it as is and we fix this later.
That being said, it makes me wonder whether we could use `subst` directly here, making it so that `blindDest` doesn't have to add the lambdas, removing the need for `error`. 

In line `322`, we'd then have: `subst (R.Arg (R.termPure ...) :> R.Arg transiAbsInput :> Id) dest`, and

src/Pirouette/Term/Transformations.hs
```diff
@@ -295,6 +296,92 @@ constrDestrId = pushCtx "constrDestrId" . rewriteM (runMaybeT . go)
...
+      let go [] _ = id
+          go (h:tl) i =
+            R.Lam (R.Ann $ fromString (s ++ show i)) h . go tl (i + 1)
+      in
+      go tys 0
+
+    -- `geneXs 3` is the representation of `[x0#2, x1#1, x2#0]`,
+    -- which are the arguments expected after a `\x0 x1 x2`.
+    geneXs :: Int -> [R.Arg ty PrtTerm]
+    geneXs n = go 0
```
VictorCMiraldo, line 368: Is there a reason to name the variables in reverse De Bruijn order? If not, then it might be easier to just define:

`geneXs n = map (\i -> R.Arg $ R.termPure (R.B (fromString $ "x" ++ show i) (toInteger $ n - 1 - i))) [0..n]`

src/Pirouette/Term/ConstraintTree.hs
```diff
@@ -150,14 +150,15 @@ pruneMaybe (Choose x ty ts) =
...
-  let (args, body) = R.getHeadLams (R.appN t $ map (R.Arg . flip R.App [] . flip R.B 0 . R.Ann . fromString) (coWithArguments opts))
-  unless (null args) $
-    logWarn $ "Executing a non-saturated term. Will use the variable names: " ++ show args
-  mctree <-
-    execute body >>= runMaybeT . if coPruneMaybe opts then pruneMaybe else return
-  logDebug $ "Translating Constraint Tree for " ++ show name
-  case mctree of
-    Nothing -> throwError' $ PEOther "termToCTree: empty tree"
-    Just tr -> return tr
+termToCTree :: (MonadPirouette m) => CTreeOpts -> Name -> PrtDef -> m (CTree Name)
```
VictorCMiraldo, line 153: nipicking it at this point, but if the function is `termToCTree`, I'd expect to receive a `PrtTerm`, not a `PrtDef`. :thinking:

VictorCMiraldo (approved):

VictorCMiraldo:
This has improved a lot, thanks @GuillaumeGen. I left a few additional small comments which I'm happy to handle on a separate PR :)

VictorCMiraldo:
closes #12

src/Pirouette/Term/ConstraintTree.hs
GuillaumeGen: Agree with you that the name should maybe be changed, but since it uses both the definition body and its type, it seems more logical to me to pass it a `PrtDef` rather than at each of the 2 use of this function do a 
```
case def of 
  DFunction _ t ty -> termToCTree opts n t ty
  _ -> throwError' "expected a function"
```

src/Pirouette/Term/Transformations.hs
GuillaumeGen: I find it more readable. I did a compromise, the order of variable name reverse of the indices, but with your syntax, much more elegant than mine.

src/Pirouette/Term/Transformations.hs
GuillaumeGen: Great idea. I first wrote it as an application when I thought at the algorithm on a piece of paper. But indeed, just writing a substitution avoids to provide dummy annotations of lambdas.

GuillaumeGen:
> closes #12

Not really, it get rids of the `pull-destr` option and of the destructor normal form (So it is already a great step forward).
However, it does not feature the nice mechanism you suggested in #12 of adding parameters to the `Next` function.

VictorCMiraldo:
I see, I forgot the parameter to `Next` was also part of #12 

This is great to be merged though, I'll go ahead and do it! Great work @GuillaumeGen ty

PR #34, opened 2021-09-07
Update plutus version

VictorCMiraldo:
Attempting to update plutus but now the IncDec contract fails with "trTerm not yet implemented type abstractions"


VictorCMiraldo:
Waiting for #36 to go in

tests/integration/GameStateMachine/Skeleton.tla
```diff
@@ -7,6 +7,18 @@ vars ==
...
+PlutusDataBounded(n) == 
+  IF n = 0
+  THEN {}
+  ELSE [ cons : {"Constr"}, arg0 : PlutusInteger, arg1 : PlutusDataBounded(n-1) ]
+
+PlutusData == UNION { PlutusDataBounded(n) : n \in 0 .. MAXDEPTH }
+
+ConstrData(n,i) == [cons |-> "Constr", arg0 |-> n, arg1 |-> i]
+MkNilData(x) == [cons |-> "Nil" ]
+
```
GuillaumeGen, line 21: It seems that `PlutusData` is a list of `Int`, but the `Nil` case is missing in the `PlutusDataBounded` function.

GuillaumeGen (approved):

tests/integration/GameStateMachine/Skeleton.tla
VictorCMiraldo: `PlutusData` is supposed to be a value of type `Data`, which is being reified as a `DefaultFun` [here](https://github.com/input-output-hk/plutus/blob/014fc0e7ca4aa8c830c651f7f1f960829ed3dfa9/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs#L92), the semantics of each of the constructors are [here](https://github.com/input-output-hk/plutus/blob/014fc0e7ca4aa8c830c651f7f1f960829ed3dfa9/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs#L357)

I decided to keep only the constructors that are needed to make the contract work for two reasons: (A) they might change the plutus `DefaultFun` again and (B) we will need to do it anyway when we write `Plutus.tla`

PR #37, opened 2021-09-09
Some initial work on defunctionalization

0xd34df00d:


src/Pirouette/Term/ToTla.hs
```diff
@@ -238,10 +276,12 @@ termToSpec opts sortedNames mainFun t = flip evalStateT tlaState0 $ flip runRead
                         >> throwError' (PEOther "CTreeIsNotAMatch")
   let next = tlaOpDef (tlaIdent "Next") [] $ tlaOr $ map (tlaIdentPrefixed "Wrapped") $ ctreeFirstMatches mctree
 
+  applies <- genApplies
+
   tlaPure $ logDebug "Assembling the spec"
   skel <- asks toSkeleton
   return $ wrapSpec skel
-         $ concat deps ++ defs ++ [next]
+         $ concat deps ++ applies ++ defs ++ [next]
```
0xd34df00d, line 284: Since, as we discussed, TLA+ is sensitive to declarations order, and definitions and `applyN` functions can depend on each other, I'm not sure about the immediately obviously correct ordering.

VictorCMiraldo: Since `applyN` are not higher-order, they can be mutually recursive:
```tla
RECURSIVE apply1(_,_)
RECURSIVE apply2(_,_,_)
...
RECURSIVE applyN(_,_,_,_, ... ,_)

apply1(f,x) == ...
apply2(f,x,y) == ...
...
applyN(f,x,y,...) == ...
```

src/Pirouette/Term/ToTla.hs
```diff
@@ -816,9 +856,23 @@ trFreeName n args = do
...
 -- |Constructs a value of a datatype as a TLA expression.
 trConstrApp :: (MonadPirouette m) => Name -> [TLA.AS_Expression] -> TlaT m TLA.AS_Expression
-trConstrApp = tlaAppName
+trConstrApp n args = mapM defun args >>= tlaAppName n
+  where
+    defun arg@(TLA.AS_QuantifierBoundFunction _ funargs _) = do
+      let arity = length funargs
+      curIdx <- saveDefun arity arg
+      pure $ defunLabel arity curIdx
+    defun arg = pure arg
```
VictorCMiraldo, line 876: That's interesting! You're doing this at the TLA+ level by saving the body of the function. I do wonder how do we handle free variables though.

Take the following program:
```haskell
main = do
  x <- read <$> getChar
  chars <- concatMap words . lines <$> readFile "whatever"
  map (+ x) chars
```

I'd expect it to be defunctionalized to something like:
```haskell
data LamCharChar = L1 Char

apply :: LamCharChar -> Char -> Char
apply (L1 x) c = c + x

map' f (x:xs) = apply f x : map' f xs

main = do
  x <- read <$> getChar
  chars <- concatMap words . lines <$> readFile "whatever"
  map' (L1 x) chars
```

Note how we keep the `x` around because it is a free variable in the body of the abstraction being defunctionalized.

src/Pirouette/Term/ToTla.hs
```diff
@@ -844,8 +898,9 @@ tlaAppNamePref pref n args = do
   arity <- case mty of
              Just ty -> return (arity $ tlaTyDropAll ty)
              Nothing -> tlaPure (logWarn $ "Undeclared type for " ++ pref ++ show (toString n)) >> return 0
-  let (tas, das) = L.splitAt arity args
-  return $ tlaFunApp (tlaOpApp (tlaIdentPrefixed pref n) tas) das
+  if arity >= length args
+  then pure $ tlaOpApp (tlaIdentPrefixed pref n) args
+  else pure $ tlaOpApp (applyFunIdent $ length args - arity) (tlaIdentPrefixed pref n : args)
```
VictorCMiraldo, line 903: Another question; 

It seems like you're dropping the application of arguments to the operator `n`. I'll try to elaborate.
Before defunctionalization, say `n` had arity 3 but it have been given 5 arguments. 
This means that `n` must have been declared as a TLA+ operator that takes 3 arguments, hence
```haskell
return $ tlaFunApp (tlaOpApp (tlaIdentPrefixed pref n) tas) das
```
would return: `prefN(x1,x2,x3)[x4, x5]`

The first three arguments are given to the TLA+ operator, and whatever left must be applied to whatever the operator returns.

VictorCMiraldo:
Looking at how this is shaping up, I start wondering whether we should do a little detour and define defunctionalization as a separate step. I can think of doing it in one of two ways:

1. Place the step _after_ we output the entire TLA+ code, which removes all occurences of `AS_QuantifierBoundFunction` and `AS_FunAp` from the program. This would reuse most of what @0xd34df00d already did, but would require some work to save free variables in the `AS_QuantifierBoundFunction` body.
2. Place the step _before_ we start translating to TLA+; hence, we would defunctionalize the System F program. Because we're not monomorphizing anything, the resulting `applyN` functions will be ill-typed. I can also think of two options here:
2.1. Just do it and declare the `applyN` function with:
    ```
    (termbind (strict)
        (vardecl apply2 (fun DefunctLbl (fun DefunctArg (fun DefunctArg DefunctRes))))
        ...)
    ```
    where `DefunctArg` and `DefunctRes` are undefined free type names. This would encode the important aspect that the `apply2` receives a label that encodes an abstraction over two arguments, but whenever we'd try to reason about its type we'd get an error saying `DefunctArg` undefined. I think this is an acceptable compromise.
2.2. Define a `type PrtTerm' = AnnTerm (Maybe PrtType) ...` representing terms with partial typing information; then define `apply2` in an untyped fashion.

I'm leaning towards favoring option `2.1` above because its the most composable; if we do it like that and sometime in the future we need to reason about `DefunctArg` (because we decide to apply a type-specific optimization, for instance), its easy to plug a monomorphizing step right before and we should end up with families of `applyN` instead.

src/Pirouette/Term/ToTla.hs
GuillaumeGen: I am not sure to understand what you are saying.
Are you stating that it would be better to write `apply2(f(x1,x2,x3),x4,x5)` rather than `apply5(f,x1,x2,x3,x4,x5)`? If so, I would be glad to hear your arguments, because I have the same feeling, but no justifications.
Or are you simply pointing that `apply5(f,x1,x2,x3,x4,x5)` is "ill-typed" in the sense that the first argument of `apply5` should not be an operator? The solution to keep this approach would be to write `apply5(curry3(f),x1,x2,x3,x4,x5)`. If someone have arguments regarding this translation, I would be glad to hear it too.

GuillaumeGen (changes requested):
It still needs some investigation to understand why the tests are failing, But the code is simple, easy to follow and goes in the good direction.

src/Pirouette/Term/ToTla.hs
GuillaumeGen: I was thinking about it while nix is updating the version of Plutus, and I think that I have a strong argument against `apply5(curry3(f),x1,x2,x3,x4,x5)`.
The definition of `curry3` would have to be `curry3(g(_,_,_)) == [x1 \in a, x2 \in b, x3\in c |-> g(x1,x2,x3)]` which would require to carry the type information somewhere and would not avoid us the rabbit hole of uselessly constructing sets with millions of elements before a simple computation.

0xd34df00d: Right, this is something I've missed, good catch. The fix seems to be easy, but do we have any contract code that'd exercise this? It's nice to have a test for this both for verifying the implementation makes sense and for future test suite.

src/Pirouette/Term/ToTla.hs
0xd34df00d: Yep, I wanted to discuss this. Would this be a problem for real-world contracts? Doing this in general requires lambda lifting, which is not the most trivial transformation out there :)

Although, on the other hand, it just occurred to me today that (especially since TLA+ is so untyped) there is a much easier route. We could just store a pair of `("label", [free vars list])` instead of just the `"label"`, which I think should work fine in this case. I'll implement this more generic behaviour.

src/Pirouette/Term/ToTla.hs
```diff
@@ -215,6 +218,123 @@ tlaWithTyVars tyvs f = do
...
+defunDtor :: TLA.AS_UnitDef -> TLA.AS_UnitDef
+defunDtor = transformBi f
+  where
+    f expr@(TLA.AS_InfixOP info TLA.AS_FunApp funExpr (TLA.AS_FunArgList _ args))
+      = case args of
+             [TLA.AS_Num {}] -> expr
+             _               -> tlaOpApp (applyFunIdent $ length args) (funExpr : args)
+    f expr = expr
+
+defunctionalize :: [TLA.AS_UnitDef] -> [TLA.AS_UnitDef]
```
VictorCMiraldo, line 330: Fantastic that you could wrap it up so neatly! :tada:

GuillaumeGen (approved):

Issue #8, opened 2021-08-06
Specialize some terms before translating to TLA.


VictorCMiraldo:
I keep on seeing small thinks like the following TLA code, which is an expression of type Bool.

```
fFoldableNil_cfoldMap(
  CConsMonoid(
    [eta169 \in Set_Bool, eta171 \in Set_Bool |-> DestBool(eta169,True,eta171)],
    False),
  LAMBDA pkh962 : pk1110 = pkh962,
  sigs957
)
```

Lets start simplifying the stub above. First, `DestBool(x, True, y)` is `if x then True else y` which is `x || y`:

```
fFoldableNil_cfoldMap(
  BoolDisjunctionMonoid,
  LAMBDA pkh962 : pk1110 = pkh962,
  sigs957
)
```

Now its easy to see what is happening! Is we translate `sigs957` as a TLA+ sequence, we could be specializing the entirety
of that code for something much more readable: 

```
\E i \in DOMAIN sigs957 : pk110 = sigs957[i]
```

In general, then, the rewrite rule is:

```
fFoldableNil_cfoldMap(BoolDisjunctionMonoid, f, xs)
  ==>
\E i \in DOMAIN xs : f(xs[i])
```

That entire expression does return a builtin TLA Boolean
With a little luck, we could be able to postpone the problem described in tweag/pirouette#3 if it only manifests itself through the use of builtin things like the example above. It's reasonable to hypothesize that user-written code wouldn't have higher order constructors a-priori.

VictorCMiraldo:
Although we're not specializing full terms just yet, @GuillaumeGen did some great work specializing certain types and their constructors in #15

PR #40, opened 2021-09-10
Specializer in the PIR syntax

GuillaumeGen:
We add a `--transfo` option, which expects a file containing rewriting rules of the form 
```
name : left-hand side ==> right-hand side
```
All those rewriting rules are applied on the system F terms, during the `processDecls` step.
This general mechanism subsumes the `cFoldMapSpecialize` and the `etaIfThenElse`, for which files representing the transformation are provided in this PR.
This is part of the #8 issue. However, it only features PIR to PIR rewriting rules, not PIR to TLA+ rules, which could be the next step.


src/Pirouette/Term/Transformations.hs
```diff
@@ -351,3 +358,176 @@ chooseHeadCase t ty args fstArg =
...
+    instantiateTy nTy m (R.TyFun a b) =
+      R.TyFun <$> instantiateTy nTy m a <*> instantiateTy nTy m b
+
+    instantiateArg :: (MonadPirouette m)
+                   => Int -> Int -> M.Map String PrtArg -> PrtArg -> m PrtArg
+    instantiateArg nTe nTy m (R.Arg t) = R.Arg <$> instantiate nTe nTy m t
+    instantiateArg nTe nTy m (R.TyArg ty) = R.TyArg <$> instantiateTy nTy m ty
+
+    isHole :: Name -> Maybe String
+    isHole x = head <$> matchRegex (mkRegex "([a-zA-Z][0-9]+)_") (T.unpack $ nameString x)
```
VictorCMiraldo, line 521: Do we need to bring in an extra dependency for that? Can't we use words starting with an underscore and check whether length x > 1 && `head x == '_'`. Or, if you rather have underscores at the end, `length x > 1 && last x == '_'`

VictorCMiraldo (commented):
Great work, thanks @GuillaumeGen !

Besides the small comments I left I was wondering whether we could use a structured format like yaml for the transformations; this way we don't need to write them in a single line, we can have a single file and we can add comments, something like:

```
etaIfThenElse:
  from: [[[{(builtin ifThenElse) A1_} x2_ ] True] False]
  to: x2_

## this transformation is important because XYZ
foldableNilEndo:
  from: ...
  to: ...
```

WDYT?

src/Pirouette/Term/Transformations.hs
GuillaumeGen: Identifiers in PIR have to start by a letter, so if we keep using the underscore, we have to put it as the last character (or any other position, but the first). I agree that it is a dependency added for a quite restricted use of regular expressions. I can modify this.

GuillaumeGen:
Yes structured files with comments are way better than the prototypic format I have chosen.

VictorCMiraldo (approved):
Awesome! Thank you @GuillaumeGen

PR #45, opened 2021-09-14
Enabling profiling

GuillaumeGen:
To enable profiling, one simply has to open a nix shell with `nix-shell --arg enableHaskellProfiling true` and then to run pirouette with RTS, like in `cabal run pirouette --enable-profiling -- --term-only tests/integration/GameStateMachine/GameStateMachine.flat +RTS -h`


VictorCMiraldo (approved):
Thanks @GuillaumeGen ; I added some profiling instructions to README.md and will force merge this now.

PR #50, opened 2022-02-04
Simplifies the nix setup by borrowing from our `plutus-libs` repo

VictorCMiraldo:


0xd34df00d (approved):
If it builds, it's good for my zero nix-fu level.

PR #53, opened 2022-02-13
Add a PlutusIR symbolic-execution engine into Pirouette

VictorCMiraldo:
Pirouette was built on the assumption that many Plutus contracts would be built with the [`StateMachine`](https://github.com/input-output-hk/plutus-apps/blob/ac3f39e1c4384ce44fdb8c321985ac447e85e411/plutus-contract/src/Plutus/Contract/StateMachine/OnChain.hs) API. From our experience, this has not been the case. Plutus contracts are being built in very diverse formats. In order to have a chance at translating these diverse contracts into TLA+, we need a full-featured symbolic execution engine.

This PR is now a collection of all the children PRs that are needed to bring a symbolic execution engine into Pirouette. This is our progress so far:

- [x] Symbolic Execution with SMT assistance of basic terms (no Higher-order, no polymorphism). (merged on: #52) 
    - [ ] Add a `sanityCheck` predicate that captures the SMT translation assumptions and gets called every time a term is being translated when running pirouette in debug mode
    - [x] Create a simpler `LanguageDef TestingLang` and add more tests to pirouette, in particular, capture the specs of symbolic execution. 
- [x] Monomorphizing PlutusIR terms as a standalone transformation ( Merged on #54 )
    - [ ] Handle de Bruijn variables instead of hardcoding `Var 1`.
- [ ] Defunctionalizing PlutusIR terms as a standalone transformation
    - [ ] Add a `sanityCheck` predicate to capture defunctionalizing assumptions.
    - [ ] Replace TLA+ defunctionalization by monomorphizing then defunctionalizing PlutusIR before translating to TLA+
- [ ] Update the relevant readmes


VictorCMiraldo:
Subsumed by #65

PR #54, opened 2022-02-16
Monomorphize WIP

0xd34df00d:
I've branched off [vcm/refactor-symb-eval](https://github.com/tweag/pirouette/tree/vcm/refactor-symb-eval) about a week ago, so the result is a mess, but it gives a rough idea. [Pirouette.Transformations.Monomorphization](/tweag/pirouette/blob/gr/monomorphize/src/Pirouette/Transformations/Monomorphization.hs) is the key module, and the rest of the changes in this PR are some very minor and non-semantic ones.

I'll rebase this on top of whatever once we decide on the merge sequence.

### DIFF
--- a/executable/Main.hs
+++ b/executable/Main.hs
@@ -63,6 +63,7 @@ import qualified Pirouette.Term.Syntax.SystemF as R
 import Pirouette.Term.ToTla
 import Pirouette.Term.Transformations
 import Pirouette.Transformations
+import Pirouette.Transformations.Monomorphization
 import qualified PlutusCore as P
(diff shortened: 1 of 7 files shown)


VictorCMiraldo (approved):
I think its looking pretty good; all that we need is some more documentation around the monomorphization module. An example or two would go a long way I believe, especially given the hardcoded "only mono tyvar index 1" constraint.

0xd34df00d:
There's still work to do wrt substitutions that we discussed, but it can be done separately.

PR #61, opened 2022-03-24
Fixing eta-expansion

GuillaumeGen:
Our eta-expansion did not distinguish well between type and term variables. I fixed it (it makes tests on eta to success).
However, I do not understand why Ormolu made such a difference in the files, I thought Ormolu was applied by default to all our commits.


VictorCMiraldo:
I've done exactly the same fix, at exactly the same time :sweat_smile: 

https://github.com/tweag/pirouette/commit/e71812548104aab146c12beea82d0d6d31eeeccf

PR #66, opened 2022-04-05
Fix monomorphization bugs

VictorCMiraldo:
This PR extends monomorphization to handle all type arguments before the first term argument; we can now fully monomorphize:

```haskell
const @Int @Bool x y
```

Which will replace the call site with `const@Int@Bool` and will produce a suitable definition.
We can't fully monomorphize:
```haskell
constWithWeirdArgs :: forall a . a -> forall b . -> b -> a

constWithWeirdArgs @Int x @Bool y
```

We will only monomoprhize the first type argument. It's unclear to me whether we want to produce a warning when we find definitions like `constWithWeirdArgs` above. Might be worthwhile for debugging purposes.

Additionally, this PR fixes important bugs in the computation of the types of monomorphized constructors and in the example parser.

### DIFF
--- a/src/Language/Pirouette/Example/Syntax.hs
+++ b/src/Language/Pirouette/Example/Syntax.hs
@@ -295,8 +295,7 @@ parseBinder binder parseVars parseBody = do
   vars <- parseVars
   guard (not $ null vars)
   b <- parseBody
-  let (n0, k0) : rest = reverse vars
-  return $ foldr (\(n, k) b' -> binder n k b') (binder n0 k0 b) rest
+  return $ foldr (uncurry binder) b vars
 
 lexeme :: Parser a -> Parser a
 lexeme = L.lexeme spaceConsumer
@@ -309,7 +308,7 @@ symbol = void . L.symbol spaceConsumer
 
 ident :: Parser String
 ident = label "identifier" $ do
-  i <- lexeme ((:) <$> lowerChar <*> many (alphaNumChar <|> char '_'))
+  i <- lexeme ((:) <$> lowerChar <*> many (alphaNumChar <|> char '_' <|> char '@'))
   guard (i `S.notMember` reservedNames)
   return i
   where
@@ -318,7 +317,7 @@ ident = label "identifier" $ do
 
 typeName :: Parser String
 typeName = label "type-identifier" $ do
-  t <- lexeme ((:) <$> upperChar <*> many (alphaNumChar <|> char '_'))
+  t <- lexeme ((:) <$> upperChar <*> many (alphaNumChar <|> char '_' <|> char '@'))
   guard (t `S.notMember` reservedTypeNames)
   return t
   where
--- a/src/Language/Pirouette/Example/Syntax.hs
+++ b/src/Language/Pirouette/Example/Syntax.hs
@@ -41,6 +41,7 @@ import qualified Pirouette.Term.Syntax.SystemF as SystF
 import Text.Megaparsec
 import Text.Megaparsec.Char
 import qualified Text.Megaparsec.Char.Lexer as L
+import Pirouette.Transformations.Utils (monoNameSep)
 
 -- * AST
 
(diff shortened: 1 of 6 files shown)


0xd34df00d (approved):
You made it much simpler than what I originally had in mind, yay!

src/Pirouette/Transformations/Monomorphization.hs
```diff
@@ -169,31 +191,54 @@ executeSpecRequest SpecRequest {origDef = HofDef {..}, ..} = M.fromList $
...
 -- higher-order _definition_ needs to be specialized (which will be handled later by 'executeSpecRequest').
 specFunApp :: forall lang. (LanguageBuiltins lang) => HOFDefs lang -> SpecFunApp lang
 specFunApp hofDefs (SystF.App (SystF.Free (TermSig name)) args)
-  | Just someDef <- name `M.lookup` hofDefs, -- We compare the entire name, not just the nameString part: x0 /= x1.
-    all isSpecArg $ take hofPolyVarsCount tyArgs = do
+  -- We compare the entire name, not just the nameString part: x0 /= x1.
+  | Just someDef <- name `M.lookup` hofDefs,
+    -- Now we ensure that there is something to specialize and that the type arguments we've
+    -- gathered are specializable arguments (ie, no bound type-variables)
+    hofPolyVarsCount >= 1,
```
serras, line 233: This is a very nitpicky thing, but I'm not a fan of patterns which relate to `where` (because I get confused by the scope). May I suggest changing it to something on the lines of the following?

```haskell
    -- with null
    not (null tyArgs),

    -- with pattern
    _ : _ <- tyArgs
```

If you then below inline `length tyArgs` in `splitArgs`, I think everything becomes more readable.

src/Pirouette/Transformations/Monomorphization.hs
```diff
@@ -211,21 +256,16 @@ specFunApp _ x = pure x
 specTyApp :: (LanguageBuiltins lang) => HOFDefs lang -> SpecTyApp lang
 specTyApp hofDefs (SystF.TyApp (SystF.Free (TySig name)) tyArgs)
   | Just someDef <- name `M.lookup` hofDefs,
-    all isSpecArg $ take hofPolyVarsCount tyArgs = do
+    hofPolyVarsCount >= 1,
```
serras, line 259: Same as above.

src/Language/Pirouette/Example/Syntax.hs
```diff
@@ -318,7 +317,7 @@ ident = label "identifier" $ do
 
 typeName :: Parser String
 typeName = label "type-identifier" $ do
-  t <- lexeme ((:) <$> upperChar <*> many (alphaNumChar <|> char '_'))
+  t <- lexeme ((:) <$> upperChar <*> many (alphaNumChar <|> char '_' <|> char '@'))
```
serras, line 320: ⚠️ WARNING: extra nitpickiness approaching.

I've been bitten in the past by trying to make identifiers too similar to full expressions. In this case, `f @Bool` is turned into `f@Bool`, and those are just too similar. May I suggest changing to a different symbol, so that we can more easily see in our code that something was a monomorphized function as opposed to a type application?

serras (approved):
Nice! I think this PR shows that the quasi-quoter for types and terms is of great help when writing tests.

serras:
> Which will replace the call site with `const@Int@Bool` and will produce a suitable definition. We can't fully monomorphize:
> 
> ```haskell
> constWithWeirdArgs :: forall a . a -> forall b . -> b -> a
> 
> constWithWeirdArgs @Int x @Bool y
> ```
> 
> We will only monomoprhize the first type argument. It's unclear to me whether we want to produce a warning when we find definitions like `constWithWeirdArgs` above. Might be worthwhile for debugging purposes.

I am right now working on a `prenexify` transformation which pushes all the "big lambdas" to the front, which would allow us to handle every rank-1 function, even when "hidden" as in that type.

src/Language/Pirouette/Example/Syntax.hs
VictorCMiraldo: We relied on `'@'` because it was what was used by monomorphization, so we needed to be able to type those names; I changed it to `monoNameSep` and defined that to be `'!'` instead.

PR #67, opened 2022-04-05
Improve CI to cache builds even if the tests fail

VictorCMiraldo:
This PR changes the pirouette CI to cache the build artifacts even if tests fail. This is done by running in two steps. First we build pirouette then test. This job succeeds if and only if pirouette builds correctly, which triggers the caches to be saved. A second job looks at the results of the tests then decides if the entire workflow passes or fails.


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,67 @@
...
+    - name: Installs nix and restores cache
+      uses: ./.github/actions/install-cached-nix
+
+    - name: Accessing the cabal cache.
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.cabal/store
+          dist-newstyle
+        key: cabal-cache-${{ github.sha }}
```
serras, line 25: Note that this refers to [the SHA of the commit that triggered the action](https://docs.github.com/es/actions/learn-github-actions/contexts#github-context). This means that the cache is not shared by different commits.

In other Haskell / Nix projects I've worked on we've usually used a hash of the files with dependencies (say, the `*.nix`, `package.yaml`, `*.cabal` files). [Here's one example](https://github.com/higherkindness/mu-haskell/blob/master/.github/workflows/deploy.yml#L27-L29) (it also used `restore-keys`, which I honestly don't know much about). This worked much better (we got faster build times), although it had the problem that sometimes the cache had to be flushed because of a huge `dist-newstyle` (`.stack-work` in our case).

VictorCMiraldo: Cache restoration first tries to look for the `key` field, then looks for the [`restore-keys`](https://github.com/tweag/pirouette/pull/67/files/1f8a48b41c61b7ca0a619bda9111fa40941dcfed#diff-bc668a2c9f2299cef15b222055b4b4d5311646caec2e7610e540cee18ae9b948R26) prefixes. This mechanism works and we've been using it in `pirouette` and `plutus-libs`.

VictorCMiraldo: That being said, I like the idea of using only the haskell files in the cache

aspiwack: The restore key is a pattern. It means that each commit will push their caches, but other commits can pull from the last cached pull.

.github/actions/install-cached-nix/action.yml
```diff
@@ -0,0 +1,37 @@
...
+        sudo mkdir -p /nix/store
+        sudo chmod -R 777 /nix
+    shell: bash
+
+  - name: Cache nix env
+    uses: actions/cache@v2
+    with:
+      path: |
+        # See https://github.com/actions/cache/pull/726
+        /nix/store/**
```
aspiwack, line 19: I don't recommend trying to cache the nix store with github actions's builtin cache. It's usually better using cachix. If you need to cache nix builds, Tweag's cachix key is available as `TWEAG_CACHIX_AUTH_TOKEN` in the Github secrets.

VictorCMiraldo: We're pulling everything from IOHK's cache anyway so I'll just skip caching the nix store altogether

PR #69, opened 2022-04-06
Prenex transformation

serras:
This transformation tries to push the type abstractions (big lambdas `/\ x : Type . `) to the front of terms, before any term abstractions (small lambdas `\ x : Int . `)

For example,

```
/\ a : Type . \ x : a . /\ b : Type . \ y : b . ...
```

is transformed to

```
/\ a : Type . /\ b : Type . \ x : a . \ y : b . ...
```

This makes it easier for later transformations to apply on type abstractions. In fact, it's common to just bail out on non-prenex types in other transformations, because those cases are quite rare and difficult to handle (usually involving some impredicative types.)

### DIFF
--- a/src/Pirouette/Transformations/Prenex.hs
+++ b/src/Pirouette/Transformations/Prenex.hs
@@ -56,8 +56,13 @@ switchLambdas :: Type lang -> Term lang -> (Type lang, Term lang)
 -- \ x . /\ a . ty --> /\ a . \x . ty
 switchLambdas (SystemF.TyFun ty1 (SystemF.TyAll annTy kindTy restOfType)) 
               (SystemF.Lam arg1 tyTm1 (SystemF.Abs annTm kindTm body))
-  = case switchLambdas (SystemF.TyFun ty1 restOfType) (SystemF.Lam arg1 tyTm1 body) of
-      (newTy, newBody) -> (SystemF.TyAll annTy kindTy newTy, SystemF.Abs annTm kindTm newBody)
+  = let -- since we are going to have an additional /\ on front,
+        -- we need to shift the bound variables by 1
+        shiftedTy1 = shiftTyBy1 0 ty1
+        shiftedTyTm1 = shiftTyBy1 0 tyTm1
+        (newTy, newBody) = switchLambdas (SystemF.TyFun shiftedTy1 restOfType)
+                           (SystemF.Lam arg1 shiftedTyTm1 body)
+    in (SystemF.TyAll annTy kindTy newTy, SystemF.Abs annTm kindTm newBody)
 -- case of small lambda, we just need to keep the invariant (see *)
 switchLambdas (SystemF.TyFun ty1 ty2) (SystemF.Lam arg1 tyTm1 body)
   = let (newTy, newBody) = switchLambdas ty2 body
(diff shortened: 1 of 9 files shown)


src/Pirouette/Transformations/Prenex.hs
```diff
@@ -0,0 +1,163 @@
...
+-- | * Prenex form of types
+-- 
+-- This transformation tries to push the type abstractions
+-- (big lambdas /\ x : Type . ) to the front of terms,
+-- before any term abstractions (small lambdas \ x : Int . )
+--
+-- For example,
+--   /\ a : Type . \ x : a . /\ b : Type . \ y : b . ...
+-- is transformed to
+--   /\ a : Type . /\ b : Type . \ x : a . \ y : b . ...
```
VictorCMiraldo, line 22:
```suggestion
-- | * Prenex form of types
-- 
-- This transformation tries to push the type abstractions
-- (big lambdas @/\ x : Type .@ ) to the front of terms,
-- before any term abstractions (small lambdas @\ x : Int .@ )
--
-- For example,
-- >  /\ a : Type . \ x : a . /\ b : Type . \ y : b . ...
-- is transformed to
-- >  /\ a : Type . /\ b : Type . \ x : a . \ y : b . ...
```

src/Pirouette/Transformations/Prenex.hs
```diff
@@ -0,0 +1,163 @@
...
+         _ -> (newTy', newBody')
+-- if we have a big lambda, just leave it
+switchLambdas (SystemF.TyAll annTy kindTy restOfType)
+              (SystemF.Abs annTm kindTm body)
+  = let (newTy, newBody) = switchLambdas restOfType body
+    in (SystemF.TyAll annTy kindTy newTy, SystemF.Abs annTm kindTm newBody)
+switchLambdas otherTy otherTm = (otherTy, otherTm)
+
+-- | Shift all the bound type variables,
+-- starting from 'from' upwards.
```
VictorCMiraldo, line 84:
```suggestion
-- starting from @from@ upwards.
```

src/Pirouette/Transformations/Prenex.hs
```diff
@@ -0,0 +1,163 @@
...
+shiftTyBy1 :: Integer -> Type lang -> Type lang
```
VictorCMiraldo, line 85: Isn't this just [`shiftCutoff from 1`](https://github.com/tweag/pirouette/blob/main/src/Pirouette/Term/Syntax/Subst.hs#L58)?

src/Pirouette/Transformations/Prenex.hs
```diff
@@ -0,0 +1,163 @@
...
+    firstTyArg (ty@SystemF.TyArg {} : rest) = Just (ty, rest)
+    firstTyArg (other : rest) = do
+      (ty, args) <- firstTyArg rest
+      pure (ty, other : args)
+
+    firstTermArg [] = Nothing
+    firstTermArg (ty@SystemF.TermArg {} : rest) = Just (ty, rest)
+    firstTermArg (other : rest) = do
+      (ty, args) <- firstTermArg rest
+      pure (ty, other : args)
```
VictorCMiraldo, line 154: These functions look generally useful and could be moved into `Pirouette.Term.Syntax.SystemF`
I'm happy with leaving them here though. :thinking:

src/Pirouette/Transformations/Prenex.hs
```diff
@@ -0,0 +1,163 @@
...
+
+-- | Apply a function only to 'FunDef's
+-- and leave the rest of definitions unchanged.
+onlyOnFunDef ::
+  (FunDef lang -> FunDef lang) ->
+  Definition lang -> Definition lang
+onlyOnFunDef f = \case
+  DFunDef funDef -> DFunDef (f funDef)
+  other          -> other
```
VictorCMiraldo, line 163: This could make its way into `Pirouette.Term.Syntax.Base`; Also, given the number of such _onFunDef_, _onDataDef_, etc.. functions, might make sense to have an issue to bring in something like `microlens` at some point.

VictorCMiraldo (approved):
Great stuff @serras, thank you! I left a few nitpicks on haddock formatting and on potentially moving some functions to different places, but I'm pre-approving to speed things up.

src/Pirouette/Transformations/Prenex.hs
serras: Not really, looking at the definition of that function:

```haskell
shiftCutoff cutoff k = subst $ foldr (\_ r -> Nothing :< r) (Inc k) [0 .. cutoff - 1]
```

There you need to say (if I understand correctly), until which de Bruijn index you want to shift, so it's an upper bound in the index. The function I've implemented shifts from that de Bruijn index to infinitive, so it works as a lower bound.

src/Pirouette/Transformations/Prenex.hs
serras: Moved in the next commit.

src/Pirouette/Transformations/Prenex.hs
VictorCMiraldo: Unrolling `shiftCutoff 2 4` gives us `subst (Nothing :< Nothing :< Inc 4)`, which means: "leave indexes 0 and 1 untouched and increment anything else by 4"

VictorCMiraldo: And, while at it, `shiftCutoff` should be called `shiftFrom` now that I think about it

serras: I've tested replacing `shiftTyBy1` with `shift` from the `Subst` module, but it doesn't give the right results.

src/Pirouette/Transformations/Prenex.hs
serras: Done in the next commit. I'm depending on `optics-core`, which is as small as `microlens`, but with better error messages.

src/Pirouette/Transformations/Prenex.hs
serras: Resolved by #73

PR #71, opened 2022-04-07
Defunctionalization tests

serras:
This PR adds tests for the defunctionalization transformation. I've used only monomorphic types, because otherwise this transformation has to be preceded by monomorphization, and things become too complicated / big.

To make everything work fine there was a lot of small changes required here and there:
- I've made the different parsers and transformations agree on the names of destructors, they are now always `match_TypeName`.
- The parser in the quasiquoter now allows `_` as the initial character of an identifier.
- Since the parser makes no effort in checking whether a definition is recursive or not, I've added an additional keyword `nonrec` which you can add to state so. This has been quite useful in the tests.

### DIFF
--- a/pirouette.cabal
+++ b/pirouette.cabal
@@ -143,6 +143,7 @@ test-suite spec
       Language.Pirouette.PlutusIR.ToTermSpec
       Pirouette.Term.Syntax.SystemFSpec
       Pirouette.Term.TransformationsSpec
+      Pirouette.Transformations.DefunctionalizationSpec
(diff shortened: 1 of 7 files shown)


VictorCMiraldo (approved):
That's really nice, thanks! Makes sense to have only monomorphic tests; we can test the whole enchilada later with some integration tests.

If it's not too involved, would be nice to have one test involving a simple datatype too, just to make sure we also cover that part of the defunctionalizer. I'm happy doing that myself later today too.

serras:
One test for data type is now in

PR #74, opened 2022-04-07
Merge different symeval implementations

serras:
The goal of this PR is to reduce the current duplication in `Pirouette.Term.Symbolic.Eval` by merging the `symeval'` and `symevalOneStep` functions into a common core, maybe with some additional parametrization.

- [X] Introduce "infinite fuel" for those cases in which symbolic evaluation should continue until the end
- [X] Merge into a single `symevalOneStep`
- [x] Add tests for symbolic evaluation
- [x] Evaluation of built-ins

### DIFF
--- a/src/Language/Pirouette/Example/Syntax.hs
+++ b/src/Language/Pirouette/Example/Syntax.hs
@@ -36,6 +36,8 @@ import qualified Data.Map as M
 import qualified Data.Set as S
 import Data.Void
 import Language.Haskell.TH.Syntax (Lift)
+import Pirouette.SMT
+import qualified Pirouette.SMT.SimpleSMT as SimpleSMT
 import Pirouette.Term.Syntax
 import qualified Pirouette.Term.Syntax.SystemF as SystF
 import Text.Megaparsec
@@ -111,6 +113,17 @@ instance LanguageBuiltinTypes Ex where
 tInt, tBool :: Type Ex
 tInt = SystF.tyPure $ SystF.Free $ TyBuiltin TyInteger
 tBool = SystF.tyPure $ SystF.Free $ TyBuiltin TyBool
+
+instance LanguageSMT Ex where
+  translateBuiltinType TyInteger = SimpleSMT.tInt
+  translateBuiltinType TyBool = SimpleSMT.tBool
+  translateBuiltinTerm TermAdd = SimpleSMT.Atom "+"
+  translateBuiltinTerm TermSub = SimpleSMT.Atom "-"
+  translateBuiltinTerm TermLt = SimpleSMT.Atom "<"
+  translateBuiltinTerm TermEq = SimpleSMT.Atom "="
+  translateBuiltinTerm TermIte = SimpleSMT.Atom "ite"
+  translateConstant (ConstInt n) = SimpleSMT.int n
+  translateConstant (ConstBool b) = SimpleSMT.bool b
 
 -- ** Syntactical Categories
 
--- a/src/Language/Pirouette/Example/Syntax.hs
+++ b/src/Language/Pirouette/Example/Syntax.hs
@@ -29,6 +29,7 @@ module Language.Pirouette.Example.Syntax where
 
 import Control.Arrow (second)
 import Control.Monad
+import Control.Monad.Except
 import Control.Monad.Combinators.Expr
 import Data.Data
 import Data.Foldable
@@ -38,6 +39,7 @@ import Data.Void
 import Language.Haskell.TH.Syntax (Lift)
 import Pirouette.SMT
 import qualified Pirouette.SMT.SimpleSMT as SimpleSMT
+import Pirouette.SMT.Translation
 import Pirouette.Term.Syntax
 import qualified Pirouette.Term.Syntax.SystemF as SystF
 import Text.Megaparsec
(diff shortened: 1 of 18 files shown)


serras:
This is a list of things which both implementations do differently, which we should clarify before continuing. The changes are marked using `diff`, and the name of the function is shortened to `se` in both cases.

## Recreating lambdas

Whereas `symeval'` returns the body applied to a new meta-variable, `symEvalOneStep` rebuilds the lambda structure around it.

```diff
se t@(R.Lam (R.Ann x) ty _) = do
  let ty' = typeFromMeta ty
  [svars] <- declSymVars [(x, ty')]
+ symeval' $ t `R.app` R.TermArg (R.App (R.Meta svars) [])
- R.Lam (R.Ann x) ty <$> symEvalOneStep (t `R.app` R.TermArg (R.App (R.Meta svars) []))
```

## Evaluating arguments for unknown heads

As above, `symEvalOneStep` always evaluates and reconstructs.

```diff
se t@(R.App hd args) = do
  mDefHead <- lift $ prtMaybeDefOf hd
  case mDefHead of
+    Nothing -> return t
-    -- If hd is not defined, we symbolically evaluate the arguments and reconstruct the term.
-    Nothing -> do
-      evaluatedArgs <- mapM (R.argMapM return symEvalOneStep) args
-      return $ R.App hd evaluatedArgs
```

## Evaluating a destructor

Here `symEvalOneStep` contains an additional check for equality which is not present in `symeval'`.

```diff
        let tyParams' = map typeFromMeta tyParams
        term' <- se term
-       if term == term'
-       then
          asum $
            for2 consList cases $ \(consName, consTy) caseTerm -> do
              let instantiatedTy = subst (foldl' (\x y -> Just y :< x) (Inc 0) tyParams') consTy
              let (consArgs, _) = R.tyFunArgs instantiatedTy
              svars <- freshSymVars consArgs
              let symbArgs = map (R.TermArg . (`R.App` []) . R.Meta) svars
              let symbCons = R.App (R.Free $ TermSig consName) (map (R.TyArg . typeToMeta) tyParams' ++ symbArgs)
              let mconstr = unify term' symbCons
              case mconstr of
                Nothing -> empty
                Just constr -> do
                  learn constr
                  return (caseTerm `R.appN` symbArgs)
-       else
-         return $ R.App hd (map R.TyArg tyParams ++ R.TermArg term' : R.TyArg tyRes : map R.TermArg cases ++ excess)
```

VictorCMiraldo:
@GuillaumeGen is probably the best to answer these questions, since he wrote the `symEvalOneStep` function

GuillaumeGen:
Here are my answers to your questions:
 1. If you do not reintroduce the lambdas, you have issues when it is an argument of a function. For instance in `map (\x -> f (x + 1)) l` where `l` is the symbolic variable, the lambda-abstraction should not be removed. I think defunctionalization should mainly mitigate the risk of it appearing, but it probably does not completely suppress it. 
 
 For the lambda at the head of the function, it is a bit less clear what the result should be, but I have the feeling that if the user did not provide names for the arguments, it is quite nice to have them bound by lambdas.
 
 2. For "unknown heads" (they are not really unknown, simply builtins in Plutus IR) then I do not know why we were not evaluating under it. It does not make sense to me to say that we do nothing to detect that `isNull l = (length l) <= (length (l++l))` is a dumb implementation of `isNull`, simply because `<=` is a builtin function.
 
 Contrary to the other ones, I was not aware that there is a difference in the implementation here.
 
 3. For the last bit, it is simply that `term' <- symeval term` means that we are fully evaluating term, and then we create branches. To replicate this behaviour with `symevalOneStep` one has to do unitary steps until there are none remaining to do, and only then create the multiple branches.
 
 The reason to fully evaluate a term before branch creations is simply that otherwise we are duplicating those steps for all the branches.

serras:
After discussing, @VictorCMiraldo and I finally realized why re-creating the lambda is necessary in some cases. However, we decided to go another route, and simply do *not* evaluate under lambdas. This is easier and we couldn't find cases in which this could make a difference.

serras:
Changing to draft because evaluation of built-ins is still something to be done.

serras:
Ready for review! At the end I've added two different ways in which built-ins can influence symbolic evaluation:

1. They can be translated into a term, as usual.
2. They may also be translated into branching.

For example, the built-in `if <c> then ... else ...` would be translated as generating two branches, one in which `<c>` holds and one in which it doesn't. This makes those built-ins as powerful as matching, and should allow us in the future to introduce language-specific information about when it's important to branch to explore the space.

src/Language/Pirouette/Example/Syntax.hs
```diff
@@ -112,6 +116,27 @@ tInt, tBool :: Type Ex
...
+  translateConstant (ConstBool b) = SimpleSMT.bool b
+  translateBuiltinTerm TermAdd [x, y] = Just $ SimpleSMT.add x y
+  translateBuiltinTerm TermSub [x, y] = Just $ SimpleSMT.sub x y
+  translateBuiltinTerm TermLt [x, y] = Just $ SimpleSMT.lt x y
+  translateBuiltinTerm TermEq [x, y] = Just $ SimpleSMT.eq x y
+  translateBuiltinTerm TermIte [_, c, t, e] = Just $ SimpleSMT.ite c t e
+  translateBuiltinTerm _ _ = Nothing
+
+instance LanguageSMTBranches Ex where
+  branchesBuiltinTerm TermIte [SystF.TyArg _, SystF.TermArg c, SystF.TermArg t, SystF.TermArg e] =
```
VictorCMiraldo, line 132: This is an interesting technique! I can see how it might be helpful for doing other interesting techniques; like teaching the SMT that `sha256` must be injective, for instance.

src/Pirouette/Term/Symbolic/Eval.hs
```diff
@@ -273,25 +352,50 @@ symeval' t@(R.App hd args) = do
...
         -- We know what is the type of all the possible term results, its whatever
         -- type we're destructing applied to its arguments, making sure it contains
         -- no meta variables.
         let tyParams' = map typeFromMeta tyParams
-        term' <- symeval term
-        asum $
+        (term', somethingWasEvaluated) <- listen $ symEvalOneStep term
+        -- We only do the case distinction if we haven't taken any step
+        -- in the previous step. Otherwise it wouldn't be a "one step" evaluator.
+        if somethingWasEvaluated == Any True
```
VictorCMiraldo, line 365: This is some funny business... :thinking: 

So, for builtins and destructors, we evaluate them in a call-by-value fashion; which might well make sense. I just wonder whether the performance will be penalized by evaluating inside an argument then reconstructing the whole term, just to have to have to traverse that term and do another step within it on the next iteration.

I don't think its worth changing it right now, but it is worth keeping it in our minds. I can see how `symEvalOneStep` could perhaps become `stmEvalSomeSteps` instead, and in those cases it just evaluates the argument a little more eagerly, perhaps until the first constructor is at the head, so pattern matching can go forward.

VictorCMiraldo (approved):
Great work, thank you @serras! I don't have many comments on a line-by-line basis. I think this is good to go and we can start exploring its use for larger scripts! :tada:

src/Language/Pirouette/Example/Syntax.hs
serras: That's exactly the plan, teaching information aside the result itself (because maybe it's not known).

src/Pirouette/Term/Symbolic/Eval.hs
serras: Indeed, there might be some exploration to be done here. The three knobs I can think of are:

- evaluating the head only as needed for pattern matching;
- whether to use a call-by-value or call-by-name strategy in application;
- whether to go into recursive operations.

PR #75, opened 2022-04-08
Use haskell-language-server 1.6.1.1

serras:
This allows us to use a more recent version of the Haskell Language Server, which brings [lots of niceties](https://github.com/haskell/haskell-language-server/releases). In the future we can keep up with the releases by simply updating the `packages.nix` file.


VictorCMiraldo (approved):
Amazing!! Thank you!

PR #77, opened 2022-04-08
Register the update to haskellNix in `sources.json` instead of hardcoding it in the nixfile

VictorCMiraldo:
This should have been a comment to #75 , by @serras, but I missed it when I first saw it. The more idiomatic way of updating a pinned dependency is using niv:

```
$ nix-shell -p niv --run "niv update haskellNix"
```

This will update the haskellNix version pinned in `nix/sources.json` for us

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -28,7 +28,7 @@ jobs:
         ## We can use the counter to invalidate the cache whenever necessary
         key: cabal-cache-0-${{ github.sha }}
         restore-keys: |
-          cabal-cache-0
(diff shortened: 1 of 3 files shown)


serras:
I tried doing that by hand (updating the commit in the `sources.json` file), but it didn't work. This is why I ended up adding those lines in the `shell.nix` script.

serras:
This business is hard :cry: Can I help somehow to get this to fruition?

VictorCMiraldo:
@serras I don't think so; I think invalidating the cache was good, now we need to get the right dependencies in there

PR #78, opened 2022-04-11
Type checker

serras:


VictorCMiraldo:
@serras, that's a great PR, thank you. It will be invaluable to be able to typecheck our transformations at each step. I pushed this through completion, which involved fixing two bugs on code that was already existing:

i) `Language.Pirouette.Example.ToTerm` was not recording the type of constructors with explicit foralls, as it should.
ii) Generation of apply functions for defunctionalized code computed the wrong return type

I'm marking this ready for review and will make a review pass myself shortly; I'd say that cosmetic work such as making the typeError messages pretty can be left as an issue

src/Language/Pirouette/Example/QuasiQuoter.hs
```diff
@@ -50,6 +58,11 @@ trQ f = case runExcept f of
   Left err -> fail err
   Right r -> return r
 
+maybeQ :: (Show e) => Either e a -> Q a
```
VictorCMiraldo, line 61: I think this could benefit from pretty-printing the type error, but that should perhaps be an issue and a separate PR later on :thinking:

src/Pirouette/Term/TypeChecker.hs
```diff
@@ -0,0 +1,280 @@
...
+
+import Data.Bifunctor
+import qualified Data.Map as Map
+import Control.Monad.Except
+import Control.Monad.Reader
+import Pirouette.Term.Syntax.Base hiding (Var, TyVar)
+import Pirouette.Term.Syntax.SystemF
+import Pirouette.Term.Syntax.Subst (shift)
+
+-- | Checks an entire set of declarations.
```
VictorCMiraldo, line 15: Should we also typecheck the main term and have a function such as: (pseudo-code below) 

```haskell
typeCheck :: (PrtReadDefs m) => m ()
typeCheck = do
  allDecls <- prtAllDefs
  mainDef <- prtMain
  case typeCheckDecls decls of
    Left err -> throwError ("Type Errors: " ++ pretty err)
    _ -> return ()
```

Now that I write that, I think that we don't have a type for main; maybe we should add a `Maybe (Type lang)` that represents the type of the main term, if any, to `PrtUnorderedDecls`.

Nevertheless, the reason I'm thinking about a function polymorphic on any `PrtReadDefs` monad is that we could
want to run the type-checker at any point in the pipeline, be it before or after sorting and removing cycles in definitions, for instance.

src/Pirouette/Term/TypeChecker.hs
```diff
@@ -0,0 +1,280 @@
...
+      flip runReaderT ((decls, []), [DeclPath name]) $
+        void $ typeCheckTerm funTy funBody
+    _ -> pure ()
+  pure ()
+
+class (LanguageBuiltins lang)
+      => LanguageBuiltinTypes lang where
+  typeOfConstant :: Constants lang    -> Type lang
+  typeOfBuiltin  :: BuiltinTerms lang -> Type lang
+  typeOfBottom   :: Type lang
```
VictorCMiraldo, line 31: Should we add this class to `Pirouette.Term.Syntax.Base`?

VictorCMiraldo (approved):
LGTM :tada: 

Some nitpicks below:

src/Language/Pirouette/Example/QuasiQuoter.hs
serras: Since only us two are working on this, I suggest to make it part of this PR. I'll take care of that tomorrow.

src/Pirouette/Term/TypeChecker.hs
serras: We definitely should! In fact, the problem was that even though the program text has a type for `main`, it was not part of `Program`, so I couldn't check it.

src/Pirouette/Term/TypeChecker.hs
serras: I guess both keeping it here (because it only applies to type checking) and moving it to base (because it's a quite common thing) make sense. I don't have a strong opinion on this.

serras: Moved there, this was nice because now some modules don't have to depend on the `TypeChecker` one.

PR #80, opened 2022-04-18
Uncovers a bug on constraint translation, crashing the SMT; Adds tests to symbolic prover

VictorCMiraldo:
I wanted to add some more involved tests to our prover and ended up finding a bug on `translateData`. While trying to diagnose the problem, I left a number of `trace` calls through the code. I found where the problem comes from, but I won't have time to solve this now, so this probably has more value as a separate PR that someone can pick up or I can continue later.

## The Bug 

Seeing the SMT crash with:
```
            Result: (error "Parse Error: <stdin>:337.79: Type ascription on constructor not satisfied, term pir_S expected sort pir_Delta but has sort pir_Nat ...pir_D pir_Delta )
```

### The Reason

We're sending the following statement to the SMT:

```
(assert (and (= pir___result ((as pir_D pir_Delta ) pir_s0 ((as pir_S pir_Delta ) ((as pir_S pir_Delta ) pir_Z ) ) ) )  ... )
```

But `D` and `S` are constructors of different types:

```
data Delta = D : Nat -> Nat -> Delta
data Nat = Z : Nat | S : Nat -> Nat
```

We should be sending the following to the SMT instead:

```
(assert (and (= pir___result ((as pir_D pir_Delta ) pir_s0 ((as pir_S pir_Nat ) ((as pir_S pir_Nat ) pir_Z ) ) ) )  ... )
```

The bug happens in the [translateData](https://github.com/tweag/pirouette/blob/dev/src/Pirouette/SMT/Constraints.hs#L192) function. Seems like my comment in there was correct, it *is* a bug! :) 

### Reproducing

The bug can be reproduced running:
```
$ cabal run spec -- -p '/prove/ && /Peano/'
```

## Orthogonal: Documentation

This PR also adds a large comment on the test inspired by the incorrectness logic paper; would be good to read that one carefully later on.

### DIFF
--- a/src/Pirouette/SMT/Constraints.hs
+++ b/src/Pirouette/SMT/Constraints.hs
@@ -20,6 +20,7 @@ import Pirouette.SMT.Translation
 import Pirouette.Term.Syntax
 import Pirouette.Term.Syntax.SystemF
 import Prettyprinter hiding (Pretty (..))
+import Debug.Trace (trace)
 
 -- TODO: this module should probably be refactored somewhere;
 -- I'm not entirely onboard with the 'translateData' funct as it is;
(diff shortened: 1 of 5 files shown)


serras (approved):
I'm going to move forward and merge this one. The fix for `translateData` should come in a different PR.

PR #83, opened 2022-04-22
TEST PR

VictorCMiraldo:
This PR is to test whether CI can properly checkout a fork or not


VictorCMiraldo:
It does; so we should stick to actions `checkout@v2.4.0` with no parameters

PR #82, opened 2022-04-21
Embed functions in Z3

serras:
The goal is to embed the functions defined by the user as uninterpreted functions in Z3.

PR #84, opened 2022-04-26
Add a very large test replicating the MinSwap issue

VictorCMiraldo:
This PR should be reviewed by first looking at the test it's adding: https://github.com/tweag/pirouette/pull/84/commits/0fbbd8dd874186f415c309e8bafe9615d3284115

Then each individual commit addresses a bug that was preventing the test to reach the symbolic engine.

### DIFF
--- a/executable/Main.hs
+++ b/executable/Main.hs
@@ -143,7 +143,7 @@ mainOpts opts uDefs = do
       else runIncorrectness (constraintFile opts) t
     symbolicExec _ _ = throwError' (PEOther "Impossible to symbolic execute a symbol which is not a function")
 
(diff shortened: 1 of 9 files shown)


VictorCMiraldo:
At this point, the example is failing because our monomorphization pass only monomorphizes polymorphic definitions that have a higher-order function as an argument. The reason for this is that we were not defining all symbols as uninterpreted functions before. Now we are, hence, we can't translate `maybeSum : forall x . Maybe x -> Maybe x -> Maybe x` to SMT because it has a `TyAll`.

Options are: either we don't transate that to SMT at all, or we instruct the monomorphizer to pick those up too.

serras:
In theory the SMT translation should be resilient to not being able to translate every single definition. In fact, right now pattern matching is not translatable, and things work fine. If the symbolic evaluation is still failing, we should check again whether the fail safe procedure for not translatable terms is working correctly.

serras:
I've agreed with @VictorCMiraldo this is going to be merged in #82, and keep working from there.

PR #91, opened 2022-05-11
Monad transformer for weighted search

serras:


src/ListT/Weighted.hs
```diff
@@ -0,0 +1,122 @@
...
+import Control.Monad.WeightedSearch (Weight (..))
+import Data.Foldable (fold, toList)
+import Data.Functor.Identity (Identity)
+
+-- | Weighted nondeterminstic computations over the weight @w@.
+data WeightedT w m a
+  = Fail
+  | Yield a (WeightedT w m a)
+  | Weight w (WeightedT w m a)
+  | Action (m (WeightedT w m a))
```
GuillaumeGen, line 35: Could you please add a comment on this datatype. My intuition of what a `Weighted n a` is, expect this type to be isomorphic to `(n, a)`.

GuillaumeGen (commented):
My impression is that `Weighted` is a type designed to do much more things than attach a weight to the objects. Changing its name is probably desirable.

src/ListT/Weighted.hs
VictorCMiraldo: I don't see `Weighted w m a` to be isomorphic to `(n , a)`; I think it is closer to a list of values, weights or a monadic value that returns lists of the same form.

package.yaml
```diff
@@ -42,6 +42,7 @@ dependencies:
   - optics-core
   - optics-th
   - ansi-terminal
+  - weighted-search
```
GuillaumeGen, line 45: I did not spot any place where this dependency is used. I assume you've added it for a future work.

src/ListT/Weighted.hs
```diff
@@ -0,0 +1,122 @@
...
+  Yield x m `mplus` n = Yield x (m `mplus` n)
+  Weight w m `mplus` Action a = Action (mplus (Weight w m) <$> a)
+  Action w `mplus` m = Action (mplus <$> w <*> pure m)
+  Weight w m `mplus` Fail = Weight w m
+  Weight w m `mplus` Yield x n = Yield x (Weight w m `mplus` n)
+  Weight w m `mplus` Weight w' n =
+    case compare w w' of
+      LT -> Weight w (m `mplus` Weight (difference w' w) n)
+      EQ -> Weight w (m `mplus` n)
+      GT -> Weight w' (Weight (difference w w') m `mplus` n)
```
VictorCMiraldo, line 90: I feel there's black magic on this instance, why don't we have a ``m `mplus` Fail = m`` clause?

package.yaml
serras: It's used for its `Weight` type class; but I could also inline that definition and remove the dependency.

VictorCMiraldo (commented):
Are we bringing a dependency just for the `Weight` typeclass? If so, is it worth it?

src/ListT/Weighted.hs
serras: I added a few comments. The idea is that `Yield` represents a `Cons`: you return an `a` and keep returning more stuff. The `Weight` is what allows you to state that one part of the computation has a different weight than other.

src/Pirouette/Term/Symbolic/Eval.hs
```diff
@@ -166,7 +166,7 @@ path x st =
 -- | A 'SymEvalT' is equivalent to a function with type:
 --
 -- > SymEvalSt lang -> SMT.Solver -> m [(a, SymEvalSt lang)]
-newtype SymEvalT lang m a = SymEvalT {symEvalT :: StateT (SymEvalSt lang) (SMT.SolverT (ListT m)) a}
+newtype SymEvalT lang m a = SymEvalT {symEvalT :: StateT (SymEvalSt lang) (SMT.SolverT (WeightedT Integer m)) a}
```
VictorCMiraldo, line 169: If this does the trick, it will be what we would call a pro-gamer move! :crossed_fingers: :astonished:

src/ListT/Weighted.hs
serras: That's the one I copied from the original package, so I don't have full insight about it yet.

I'm pretty sure that line you mention could be added; but I'm a bit concerned about potential loss of laziness in the case in which we `Yield` a nested amount of elements, since in that case the current definition of `mplus` will never look at the other side until we have consumed all those elements.

serras:
> Are we bringing a dependency just for the `Weight` typeclass? If so, is it worth it?

Maybe not. But on the other hand the package is quite lightweight and doesn't bring any additional dependency on its own.

src/ListT/Weighted.hs
VictorCMiraldo: So I guess that `Weighted w m a` is related to `m [a]` through permutation of the list elements (and blissful ignorance of the monadic part...)

src/Pirouette/Term/Symbolic/Eval.hs
serras: Everything still works, but there's no usage of the weights yet, so the results are the same (and the performance suffers a bit because before we were using lists).

VictorCMiraldo:
> Maybe not. But on the other hand the package is quite lightweight and doesn't bring any additional dependency on its own.

It's also not a bad idea to just hardcode the weights to be `Integer` and not have an additional type-variable

src/ListT/Weighted.hs
serras: Exactly. The key point is this line in the definition of `mplus`:

```haskell
  Weight w m `mplus` Yield x n = Yield x (Weight w m `mplus` n)
```

which says that "if you have something with a weight on one side, but can yield on the other, just do so".

VictorCMiraldo:
Seems to work, according to CI! :tada: 

Yet, the MinSwap example went from 2.8s to run to closer to 9s... this might be a consequence of a number of things or just a coincidence that our previous depth-first search would hit the bug faster

serras:
Weights haven't been put in use yet, so this really means that we are slower in general. Hopefully we can combat this with a more intelligent search strategy.

src/ListT/Weighted.hs
```diff
@@ -0,0 +1,155 @@
...
+
+-- | Weighted nondeterminstic computations over the weight @w@.
+data WeightedT w m a
+  = Fail
+  | -- | returns one value, and then more of them
+    Yield a (WeightedT w m a)
+  | -- | signals that the internal computation is weighted
+    Weight w (WeightedT w m a)
+  | -- | perform an action
+    Action (m (WeightedT w m a))
```
VictorCMiraldo, line 41: It seems we could also define this as:
```suggestion
data WeightedT w m a
  = Empty
  | -- | returns one value, and then more of them
    Yield (m a) (m (WeightedT w m a))
  | -- | signals that the internal computation is weighted
    Weight w (WeightedT w m a)
```
I wonder about the tradeoffs :thinking:

src/ListT/Weighted.hs
GuillaumeGen: I can read the code and understand it. But when there is a function of type `f :: Weighted Int a -> Weighted Int [a]` what I guess when reading this type is that `f` takes an object of type `x` (with a weight `w`) and outputs a list of such objects (once again with a weight). It is not at all the meaning of the `Weighted` monad. And I think that it is the kind of thing which makes code hard to read.

src/ListT/Weighted.hs
VictorCMiraldo: That's a very good concern!

src/ListT/Weighted.hs
GuillaumeGen: I realize I probably badly expressed myself the first time. I did not intended to say that the type `Weighted w a` you wrote looks isomorphic to `(w, a)`. I wanted to state that if I jump in a new code and see that something is of type `Weighted w a`, I would expect it to be isomorphic to `(w, a)`.

serras: Is this mostly a naming problem? If I renamed it to `WeightedSearch w a`, would that give more the idea that this is like a list monad but with weights?

src/ListT/Weighted.hs
serras: I started with that, but I didn't manage to implement `mplus`. So I turned into this other representation.

src/ListT/Weighted.hs
VictorCMiraldo: But in the case ``Action m `mplus` Yield x n`` we're not yielding before executing the action, even though we could.

src/ListT/Weighted.hs
GuillaumeGen: What you are doing is (isomorphic to) a list of `(w, a)` with the guarantee (at type-level) that this list is sorted on the "weights" (ie the first elements). I think that `WeightedList` or `WeightedSet` would convey the right intuition about the fact that we are dealing with a bag of elements, with the guarantee that consumption would be done by increasing weights.

VictorCMiraldo: Following the denotation of `Weighted w a` as a set of `a`s, reasing the type of `f :: Weighted Int a -> Weighted Int [a]` is simple: it takes a set of `a`s and returns a set of list of `a`s. The `a`s themselves have no weights associated to them, the name is a trap! All the weights do is fix the order in which they are returned (I picked a set as a denotation here precisely to ignore that)

src/ListT/Weighted.hs
GuillaumeGen: I think that this change the semantic.  ```Action Nothing `mplus` Yield x n``` currently outputs `Action Nothing`, with this change it would outputs `Yield x (Action Nothing)` (assuming `n` is not a `Yield` itself).

src/ListT/Weighted.hs
GuillaumeGen: Yes, it is a naming issue mainly.

src/ListT/Weighted.hs
serras: Mmmm... after thinking a bit more, I think that indeed `mplus (Action m) thing` should always first execute the `action` before `mplus`ing with `thing`. So I'm leaving it as it was.

src/ListT/Weighted.hs
serras: Renamed in the latest commit, not it's `WeightedListT`.

src/ListT/Weighted.hs
VictorCMiraldo: Fair! All of this work because we want to burry an IO monad somewhere. Makes me wonder whether we should just `unsafePerformIO` the heck out of the SMT interface

VictorCMiraldo: This typechecks:

```haskell
instance (Alternative m) => Alternative (T m) where
  Fail <|> r = r
  r <|> Fail = r
  Yield ma mr <|> r = Yield ma (flip (<|>) r <$> mr)
  Weight w r <|> Yield mb ms = Yield mb ((Weight w r <|>) <$> ms)
  Weight w r <|> Weight w' s =
    case compare w w' of
      EQ -> Weight w  (r <|> s)
      LT -> Weight w  (r <|> Weight (w' - w) s)
      GT -> Weight w' (r <|> Weight (w - w') s)
  ```

The advantage I see here is that it also orders the execution of effects; our current representation with `Action` only orders the results, but still executes the actions in the order it encounters them, no? I could be totally wrong though, so don't quote me on that!

File: src/ListT/Weighted.hs (hunk 3 above)
Comment (VictorCMiraldo) [reply src/ListT/Weighted.hs]: > I think that this change the semantic. Action Nothing `mplus` Yield x n currently outputs Action Nothing, with this change it would outputs Yield x (Action Nothing) (assuming n is not a Yield itself).

I don't think so... say: ``Action Nothing `mplus` Yield x n :: WeightedT Maybe ()`, the denotation is at `Maybe [()]`,
```
denot (Action Nothing) == Nothing
denot (Yield x (Action Nothing)) == (x :) <$> denot (Action Nothing) == Nothing
```

What it changes is the order in which effects get executed

File: src/ListT/Weighted.hs (hunk 5 above)
Comment (serras) [reply src/ListT/Weighted.hs]: Now I cannot get to implement `>>=` :thinking:  The problem is that we have:

```haskell
Yield x m >>= f = mplus (f x) <$> ((>>= f) <$> m)
```

But how do I know whether I have a `Yield` or something else to put on the top of that?

VictorCMiraldo: hmm... Maybe we play the `ListT` trick:

```haskell
data T m a
  = Yield (m (a, T m a))
  | Weight Natural (T m a)
```

I'd assume that now we can copy the bind from `ListT`

VictorCMiraldo: But now I think this is the same as you've got...

VictorCMiraldo (approved):
Looks great to me!

PR #92, opened 2022-05-13
Generalized stopping conditions for symbolic evaluation

serras:
This is a quick generalization of the usage of "fuel" to stop symbolic evaluation. The idea is
1. Gather different statistics during evaluation (fuel used, # of constructors introduced, etcetera);
2. The "runner" function take a new `StoppingCondition` which is nothing more than a predicate `SymEvalStatistics -> Bool`.

So now our previous runners become:
- Infinite runners simply use `const False` as `StoppingCondition`,
- Runners with limited fuel are implemented as `\st -> sestConsumedFuel st > availableFuel`,
- Runners limited by constructors can now be implemented as `\st -> sestConstructors st > maxConstructors`.

### DIFF
--- a/executable/Main.hs
+++ b/executable/Main.hs
@@ -34,7 +34,7 @@ import Pirouette.Monad.Logger
 import Language.Pirouette.PlutusIR.SMT ()
 import Language.Pirouette.PlutusIR.Builtins
 import Language.Pirouette.PlutusIR.ToTerm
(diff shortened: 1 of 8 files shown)


VictorCMiraldo (approved):
Pardon my delay in reviewing this; it does look like a reasonable approach for now

PR #93, opened 2022-05-19
Remove executable and make `incorrectnessLogic` a tasty assertion

florentc:
This is the beginning of the move from pirouette as an app to pirouette as an API. Right now, all the exec has been deleted. Please require which feature you would like re-exposed in the new top-level `Pirouette` module.

As for `incorrectnessLogic`, it is now a tasty unit test assertion which fails in case a counterexample is found or an unexpected error occurs. The existing color-coded pretty-printing on stdout has been kept for now for convenience. In the future we might want to clean things up to keep test trees clean, or put in in the test log.


VictorCMiraldo:
Why not clean it out now? We can always split it out in two different functions:
```
runIncorrectnessLogic :: ... -> IO IncorrectnessResult

assertIL :: Assertion
assertIL = do
  x <- runIncorrectnessLogic
  case x of
    Ok -> return ()
    _ -> assertionFailure "some reason"

incorrectnessLogicRepl :: ... -> IO ()
incorrectnessLogicRepl = runIncorrectnessLogic >>= putStrLn . prettyWithColors
```

src/Pirouette/Term/Symbolic/Prover/Runner.hs
```diff
@@ -23,25 +24,33 @@ data AssumeProve lang = Term lang :==>: Term lang
 type IncorrectnessResult lang =
   Either String (Maybe (Path lang (EvaluationWitness lang)))
 
+-- For now only out of fuel stopping condition
+newtype StoppingCondition = StopOutOfFuel Int
+
+-- | Parameters for an incorrectness logic run
+data ILParams lang = ILParams
```
VictorCMiraldo, line 31: To remain consistent with the other functions, maybe we should call this type `IncorrectnessLogicParams`; I'm fine with using the `il` prefix for the constructor and its fields.

VictorCMiraldo (approved):
Looks good, thank yo @florentc ! One step closer to "release"

PR #98, opened 2022-05-27
Be more precise with symbolic evaluation of arguments

serras:


src/Pirouette/Term/Symbolic/Eval.hs
```diff
@@ -563,9 +562,16 @@ symEvalOneStep t@(R.App hd args) = case hd of
   -- in any other case don't try too hard
   _ -> justEvaluateArgs
   where
```
lucaspena, line 564: Maybe add a comment here stating that we only evaluate the first argument (if it can be evaluated), instead of evaluating all arguments in parallel as before.

serras:
This has been superseded by #100. Marking as draft to prevent merging; this should be eventually closed once #100 lands.

serras:
Closed because #100 has been merged

Issue #94, opened 2022-05-20
In `dev`: Test exhaustiveness of symbolic evaluator


VictorCMiraldo:
We need a test that the symbolic evaluator really does check all the possible values of a type up to a certain amount of constructors; something like checking that symbolic evaluating with stopping condition `sestConstructors st < 20` the term `\(x : Pair (List Int) (List Int)) -> product (fst x) + product (snd x)` will explore branches for ``x `elem` [ (as, bs) | length as + length bs == 19 ]` (the pair has one constructor)

serras:
Note that the actual statistic right now is how unifications are done during `match`es are found during symbolic execution.

VictorCMiraldo:
Is this why we're doing #98 ? If so, does it make sense to make add another dial to pirouette or does it make sense to refine how statistics are computed and *remove* a dial?

serras:
I had a discussion with @lucaspena, and the problem seems more complex that it appears at first, since there are a lot of traps which would make `f e1 e2` just recurse forever on `e1` if we don't have a cut-off. To prevent that, the code before #98 was:

1. Evaluating one step of all arguments in parallel, and only moving to the function body if none of those steps resulted in further evaluation;
2. Evaluating the body, the "assume" and the "to prove" functions in parallel. This is really important, as you want to prune the search as early as possible using the "assume" function.

Since the counter is only checked once after taking those parallel steps, it could be the case that you've seen several `match`es and thus the counter of known constructors is reached.

#98 was a simple attempt at being more precise by using a depth-first approach to the evaluation of arguments. We only look at argument 2 if argument 1 didn't take any further steps. This risks going into an infinite loop (imagine we symbolically evaluation something on a list, we get larger and larger symbolic lists at each step); but it's safe if you have a cut-off, since at some point you'll run out of fuel/constructors/... and the stopping condition for that branch would be true.

Thinking about it this weekend I've come to another possible strategy to be at the same time (1) more precise and (2) not risk into unfairness. Here's the strategy applied to a list of expressions:

1. Traverse the list looking for a `match`, and if found, take a step on that one (which introduces constructors);
2. If none of them is a `match`, then take a step in parallel on every expression.

I'm going to work on that on a separate branch.

VictorCMiraldo:
@lucaspena, do you have some work on this that should become a PR? Please coordinate with @facundominguez so we don't duplicate work. He will be picking this up as part of his pirouette onboarding

facundominguez:
[...]

PR #100, opened 2022-05-30
Evaluate destructors earlier than other arguments

serras:
See the discussion at https://github.com/tweag/pirouette/issues/94#issuecomment-1140823333


VictorCMiraldo (approved):
Looks good to me, thanks @serras! It is interesting that the `isUnity` example now takes 10 seconds to complete. I have no idea about why that is, this change seems harmless enough

VictorCMiraldo:
The increase in runtime makes sense; before, say we had three arguments `[x, y ,z]`, we would expand all three then run `prune`. Now, we expand one at a time (if they're all `match`es, for instance) and run `prune` multiple times.

Issue #101, opened 2022-06-01
In `dev`: support `PlutusIR` type synonyms


florentc:
Validators in PlutusIR turn out to make use of type synonyms.
https://github.com/tweag/pirouette/blob/dev/src/Language/Pirouette/PlutusIR/ToTerm.hs#L205
Handling those is necessary to progress on #88.

VictorCMiraldo:
Closed by #103

Issue #88, opened 2022-05-10
in `dev`: support as much of `PlutusIR` primitives as possible


VictorCMiraldo:
Currently, we can only run pirouette over its `Pirouette.Language.Example`, we should go ahead and implement all the necessary classes for `PlutusIR`; these are:

- [x] `LanguageBuiltinTypes`: https://github.com/tweag/pirouette/blob/9b83f80f0281bf08df94bedb3122c8348e84938a/src/Language/Pirouette/Example/Syntax.hs#L120
- [x] `LanguageSMT`: https://github.com/tweag/pirouette/blob/9b83f80f0281bf08df94bedb3122c8348e84938a/src/Language/Pirouette/Example/Syntax.hs#L145
- [x] `LaguageSMTBranches`: https://github.com/tweag/pirouette/blob/9b83f80f0281bf08df94bedb3122c8348e84938a/src/Language/Pirouette/Example/Syntax.hs#L193

Additionally, we should also have a quasi-quoter for writing the proof obligations in `PlutusIR`; this can probably be reasonably easily done through some surgery on our example language syntax module; I reckon we can parameterize that AST and parser by a choice (and parser) for builtins and use that to quasi-quote `PlutusIR` terms.

VictorCMiraldo:
@florentc , I updated this issue with some comments and more pointers

VictorCMiraldo:
Another important aspect to this issue is the "as much of `PlutusIR` **as possible**"... The set of primitives is too large to be handled on a single go. I'd suggest we gather two client contracts of your choosing, produce their respective flat files and save them as _golden_ tests. As long as we can handle those we're fine. We should slowly increase the set of our golden tests.

VictorCMiraldo:
FWIW, Plutus primitives are here: https://github.com/input-output-hk/plutus/blob/master/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs

VictorCMiraldo:
#102 has the quasiquoter and is able to typecheck two golden pir files.

VictorCMiraldo:
With the merging of #113, I'll consider this closed. We will certainly be finding more examples that won't work and will require the addition of some builtin, but that's an issue for the future.

PR #102, opened 2022-06-01
Typecheck two PIR files

florentc:
This PR adds partial support for Plutus IR primitives to progress on #88. Development is guided for now by running pirouette on the split and auction validators.

- typing: this PR provides typing for the constants, and a subset of builtins including those used by the auction validator. Testing is not possible until #101 is solved.
- conversion to SMT lib: this provides support for basic builtins on integers and strings, enough for the split validator. The remaining translations for builtins used by auction (lists and data) are listed and remain to be implemented

------
edit by @VictorCMiraldo:

- This PR was downsized to simply typechecking and symbolically executing two golden PIR files (`auction.flat` and `split.flat`) that come from our `plutus-libs` repo
- Moreover, it also handles the type-dependencies of let-bindings floated to the top level


VictorCMiraldo:
I just merged #103 into `dev`, then `dev` into here and I could see the type-checking problems, which shows that the types of certain builtins is probably wrong. This PR is good to go as soon as the tests pass; I'll mark it as ready for review.

VictorCMiraldo:
We currently have a bug w.r.t the translation of dependencies of terms brought from within a let statement; check the definition of `go2339`:  

```
go2339 |->
  2#k2300 -> (2#k2300 -> 2#k2300 -> Bool1980) -> 1#v2301 -> (List2008 (Tuple22015 2#k2300
                                                                                  0#r2302)) -> (These2212 1#v2301
                                                                                                          0#r2302)
```

It references a number of bound type variables, but these are not being properly declared: it should read something like:

```
go2339 |->
  ∀ (k2300 : *) (v2301 : *) (r2302 : *)
  2#k2300 -> (2#k2300 -> 2#k2300 -> Bool1980) -> 1#v2301 -> (List2008 (Tuple22015 2#k2300
                                                                                  0#r2302)) -> (These2212 1#v2301
                                                                                                          0#r2302)
 ```

VictorCMiraldo:
The bug was present before #103 was merged

VictorCMiraldo:
The two golden tests for typechecking PIR files are now passing; this PR turned out to be quite important and implemented handling of type dependencies when floating let-bindings to the top-level. Moreover, it also removed the types of all unused PlutusIR primitives. This is because a number of the types that we guessed before were wrong; if we don't know a type, we should throw an error so the user of pirouette doesn't spend time debugging elsewhere.

VictorCMiraldo:
That's pretty confusing! Yet, I think the lines you send me are not the type of `chooseList` and `chooseData`, they're an excerpt of System F exemplifying how to use those primitives. In the body of the terms you linked, `chooseList` and `chooseData` have types which are consistent with what we currently have on our `typeOfBuiltin`

VictorCMiraldo:
I think this PR is in good enough state to be reviewed then merged. Current state of the art is:

- We should now be able write PlutusIR terms with our quasiquoter: `[pir| \x : Integer . x + 1 |]`, although the parser is incomplete and this has not been tested just yet.
- We can type two golden PIR files

PR #103, opened 2022-06-03
Process PlutusIR type synonyms

florentc:
Incomplete

This is aimed at solving #101 

This PR adds type synonyms to the environment (`Env`) of the translation context (`TrM`) of PlutusIR programs.
Resolution of type synonyms is implemented in `trType` as an alternative to the existing translation in case the type variable name is found in the environment.

Populating that environment is the more tricky remaining part.

`Env` is carried by a `Reader` monad in `TrM` but the translations of bindings `trDataOrTypeBinding` are applied sequentially in `trBindings`.

`mapM` (at line 118) is replaced by an iteration through the list of bindings that adds previous type synonyms declarations in the `Env` for each `trDataOrTypeBinding` (see `mapMTypeSynonyms`). The type synonyms field of the environment is non-empty only when translating those individual bindings.

Right now, there is a typing error in `mapMTypeSynonyms` due to mixes of `name` `tyname` and concrete `Name` in definitions of `TrM`, `Env`, and `trBindings`. Some discrepencies seem to be unintended design or anwser specific needs that may have to be taken into account before undergoing too big refactoring.


VictorCMiraldo:
> `Env` is carried by a `Reader` monad in `TrM` but the translations of bindings `trDataOrTypeBinding` are applied sequentially in `trBindings`

I think that carrying the type-synonym dictionary will be easier if its done in the `StateT` part of `TrM`; is there a particular reason for choosing to do in the `ReaderT` layer?

VictorCMiraldo:
@florentc, did you get to see my comment above?

VictorCMiraldo (approved):

PR #104, opened 2022-06-08
Update nixpkgs to release-22.05

serras:
⚠️ After this PR is merged, developers of Pirouette may need to clean (parts of) their `~/.cabal` folder.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -64,11 +64,8 @@ jobs:
       run: |
         echo "!! CABAL TEST OUTPUT:"
         cat pirouette-cabal-test.out
-        echo "!! HLINT OUTPUT:"
-        cat pirouette-hlint.out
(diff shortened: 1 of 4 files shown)


VictorCMiraldo (approved):
LGTM! Thanks @serras! Our new `hlint` is being annoying on a few places. I'm starting to think we should just disable `hlint` at some point...

serras:
I agree, HLint is now integrated in HLS, so I think most of us already apply the useful suggestions.

Something similar happens with Ormolu: what about turning it into a pre-commit hook, so that everybody runs it before commiting? Then we won't get annoying "your build failed because of a missing newline" errors.

VictorCMiraldo:
I have ormolu set up as a pre-commit hook, there are instructions here: https://github.com/tweag/plutus-libs#pre-commit-hooks-and-ci

AFAIK pre-commit hooks have to be installed locally... Maybe we could pull off some nix magic?

PR #108, opened 2022-06-15
First steps on symbolic eval of Plutus IR

serras:


serras:
Reminder for future me: use `declare-datatypes` (as described in http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2021-05-12.pdf) to declare all data types at once. That allow us to work around any potential recursion or dependency between them.

src/Pirouette/Transformations/Contextualize.hs
```diff
@@ -0,0 +1,81 @@
...
+import qualified Data.Map as Map
+import Pirouette.Monad
+import Pirouette.Term.Syntax as S
+import Pirouette.Term.Syntax.SystemF as SystF
+
+-- | Make all the 'Name's in the term with empty uniques
+-- refer to those in the context.
+-- This is especially useful when working with lonely terms
+-- created by the quasiquoter, but which ultimately need
+-- to refer back to elements in a 'Program'.
```
VictorCMiraldo, line 13:
```suggestion
-- to refer back to elements in a 'Program'. In particular, the `nameUnique` part might need adjustments.
-- Say you called the quasiquoter with @[pir| \x : Integer -> fib x |]@, you'd want `fib` to refers to something 
-- from the context. The `fib` in the context will actually be called @Name "fib" (Just u)@, where `u` is a 
-- unique identifier but the `fib` that came from the quasiquoter will be @Name "fib" Nothing@, hence we
-- need to adjust the 'nameUnique' portions.
```

VictorCMiraldo: It took me a long time to realize why we even need this; I'd suggest appending some information to the comment

tests/unit/Language/Pirouette/PlutusIR/SymEvalSpec.hs
```diff
@@ -0,0 +1,70 @@
...
+    assume' <- contextualizeTerm assume
+    toProve' <- contextualizeTerm toProve
+    toDo (Problem tyRes' fn' assume' toProve')
+
+tests :: [TestTree]
+tests =
+  [ testGroup
+      "simple triples"
+      [ expectFail $ testCase "just evaluation" $
+          execFromPIRFile proveUnbounded
```
VictorCMiraldo, line 62: This looks very nice! :tada:

VictorCMiraldo (approved):
LGTM

Issue #106, opened 2022-06-14
In `dev`: Remove `Logger` and rename `LanguageSMTBranches`


VictorCMiraldo:
Check: https://github.com/tweag/pirouette/pull/99#issuecomment-1155045558

VictorCMiraldo:
Additionally, rename and document `LanguageSMTBranches` according to https://github.com/tweag/pirouette/pull/99#discussion_r897627279

PR #109, opened 2022-06-16
Remove `Logger` monad and remove `PirouetteBase` constraint

VictorCMiraldo:
This PR addresses #106; it:

- [x] Removes the unused `Logger` monad, as a consequence, it also removes the unused `MonadError` constraint and instead just throws `error`s; the errors we had were never caught anywhere anyway, lets deal with adding more monads only if we need them in the future.
- [x] Renames `LanguageSMTBranches` to `LanguageSymEval`


src/Pirouette/SMT/FromTerm.hs
```diff
@@ -125,17 +104,14 @@ translateTerm knownNames (Raw.App var args) = case var of
       Just t -> return t
   Raw.Free (TermSig name) -> do
     _ <- traceMe ("translateApp: " ++ show name) (return ())
-    defn <- liftTranslator ("translateApp: Unknown name: " <> show name) (prtDefOf name)
+    defn <- lift $ lift $ prtDefOf name
```
0xd34df00d, line 107: Very curious `lift $ lift` is not equivalent to `lift` in this context, but the type error is not too obvious at a first glance.

0xd34df00d (approved):
Lots of code removed indeed, and that's always a good thing!

src/Language/Pirouette/Example.hs
```diff
@@ -235,9 +236,25 @@ isConstant (BConstant _) = True
...
+--
+-- Say we translate the builtin @if@ statement to its SMT counterpart and evaluate:
+--
+-- > (assert (= x (ifthenelse (isEven 5) 42 0)))
+-- > (assert (= x 0))
+-- > (check-sat)
+--
+-- The SMT will say the above is unsat, and will give a model that assumes @isEven 5@ to be true,
+-- yielding @x = 42@. That's because @isEven@ is an uninterpreted function, so the SMT has no information
+-- about it.
```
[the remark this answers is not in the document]
serras: I agree, this looks like wonderful information for any implementor of the class.

src/Pirouette/Monad.hs
```diff
@@ -57,57 +51,40 @@ data PrtError
...
--- | The base monadic layer we use in most places.
-type PirouetteBase m = (MonadError PrtErrorCtx m, MonadLogger m, MonadFail m)
-
--- | 'throwError' variant that automatically gets the logger context.
-throwError' :: (PirouetteBase m) => PrtError -> m a
-throwError' msg = do
-  err <- flip PrtErrorCtx msg <$> context
-  throwError err
+prtError :: PrtError -> a
+prtError = error . show
```
serras, line 55: What is the reasoning for moving from `MonadFail` to `error`? I'm on the fence here: I like the removal of `MonadError`, but I feel like `error` is too difficult to catch.

src/Pirouette/SMT/FromTerm.hs
serras: Maybe this is the monomorphic `lift` which only "goes up one step"?

src/Pirouette/Symbolic/Eval/Types.hs
```diff
@@ -1,24 +1,56 @@
...
 import Pirouette.Term.Syntax
 import Prettyprinter hiding (Pretty (..))
-import qualified Data.Set as S
-
--- import Debug.Trace (trace)
+import qualified PureSMT
+
+-- | The 'LanguageSymEval' class is used to instruct the symbolic evaluator on how to branch on certain builtins.
+-- It is inherently different from 'SMT.LanguageSMT' in the sense that, for instance, the 'SMT.LanguageSMT'
+-- translation of a primitive @if@ statement might use the @ifthenelse@ SMT primitive. However, the
```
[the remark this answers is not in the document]
VictorCMiraldo: I mean, the translation makes sense, it's just that for the purposes of symbolically evaluating, we can't rely on it. 
If we ever write different analyses in pirouette, we might want to translate if statements into SMT.

src/Pirouette/Transformations.hs
```diff
@@ -23,8 +23,8 @@ import Pirouette.Term.TransitiveDeps
 -- | Removes all Even-Odd mutually recursive functions from the program.
 --  When successfull, sets the 'tord' state field with a list of names indicating the order in which
 --  they should be defined so that the dependencies of a term @T@ are defined before @T@.
-elimEvenOddMutRec :: forall lang m. (LanguageBuiltins lang, PirouetteBase m) => PrtUnorderedDefs lang -> m (PrtOrderedDefs lang)
-elimEvenOddMutRec udefs = do
+elimEvenOddMutRec :: forall lang. (LanguageBuiltins lang) => PrtUnorderedDefs lang -> PrtOrderedDefs lang
+elimEvenOddMutRec udefs = runIdentity $ do
```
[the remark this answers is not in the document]
VictorCMiraldo: That's why I decided to keep it; let's just leave the functions in the where clause monadic so we don't have to worry too much. The toplevel declaration is pure, and that's what matters :)

tests/unit/Language/Pirouette/PlutusIR/ToTermSpec.hs
```diff
@@ -58,7 +57,7 @@ tests = (:[]) $ withResource acquire (const $ return ()) $ \progAct ->
         let shortN = head $ filter ((== "short") . nameString) $ M.keys decls
         let (DFunction _ long _) = fromJust $ M.lookup longN decls
         let (DFunction _ short _) = fromJust $ M.lookup shortN decls
-        mockPrt (runReaderT (expandDefs long) (mocked decls)) @?= Right short
+        runIdentity (runReaderT (expandDefs long) (mocked decls)) @?= short
```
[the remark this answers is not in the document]
VictorCMiraldo: You are corrrect! Thanks!

src/Pirouette/Symbolic/Eval/Types.hs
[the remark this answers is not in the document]
serras: One possibility is what was done before with Plutus IR: simply lowering the whole program into a big SMT constraint.

src/Pirouette/SMT/FromTerm.hs
VictorCMiraldo: I think it's because the monad in question is a `ReaderT x (ReaderT y)`, so lift gets confused

src/Language/Pirouette/Example.hs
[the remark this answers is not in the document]
VictorCMiraldo: Makes sense; agreed! Will move

src/Pirouette/Symbolic/Eval/Types.hs
```diff
@@ -1,24 +1,56 @@
...
+-- >   |
+-- > LanguageSMT --> defines translation of built-ins into SMT
+-- >   |             (not every term can be translated)
+-- >   |
+-- > LanguageSymEval --> defines at which points the symbolic evaluator has to do sth. special
+class (SMT.LanguageSMT lang) => LanguageSymEval lang where
+  -- | Injection of different cases in the symbolic evaluator.
+  -- For example, one can introduce a 'if_then_else' built-in
+  -- and implement this method to look at both possibilities.
+  branchesBuiltinTerm ::
```
serras, line 47: Maybe some info about the arguments would help maintainability here. Since I wrote the function, here are the arguments:

```haskell
    (SMT.ToSMT meta, PirouetteReadDefs lang m) =>
    -- | Head of the application to translate
    BuiltinTerms lang ->
    -- | Translation function to SMT, in case you need to recursively translate an argument
    (TermMeta lang meta -> m (Maybe PureSMT.SExpr)) ->
    -- | Arguments of the application to translate
    [ArgMeta lang meta] ->
    -- | If `Nothing`, this means that this built-in term
    -- does not require anything special from the
    -- symbolic evaluator. For example, it might be
    -- that it's translated fine by 'LanguageSMT'.
    -- If `Just`, it provides information of what to
    -- assume in each branch and the term to 
    -- symbolically evaluate further.
    m (Maybe [SMT.Branch lang meta])
```

serras (approved):

src/Pirouette/Monad.hs
VictorCMiraldo: The reasoning is that we were never catching any of those errors anyway. They were all un-recoverable errors. Things like: "Symbol not defined" is not something we can recover from.

src/Pirouette/Symbolic/Eval/Types.hs
VictorCMiraldo: Awesome! Thanks! :)

PR #110, opened 2022-06-16
Use pure `solve` function

VictorCMiraldo:
Using a `pure` solve function and parallel symbolic evaluation seems promising; our state monad is hurting us a little, I might work on removing that state monad if not extremely necessary. 

Also, another curious fact is that keeping the lock in `pushMStack` and `popMStack` causes a deadlock, but removing it makes everything go through. I'm a bit puzzled by it.

### DIFF
--- a/src/Pirouette/Symbolic/Eval.hs
+++ b/src/Pirouette/Symbolic/Eval.hs
@@ -193,27 +193,21 @@ instance (SymEvalConstr lang) => Alternative (SymEval lang) where
       runSymEvalRaw solvers defs st xs
         <|> runSymEvalRaw solvers defs st ys
 
-{-
 -- | Prune the set of paths in the current set.
-prune :: forall lang m a. (SymEvalConstr lang m) => SymEval lang m a -> SymEval lang m a
+prune :: forall lang a. (SymEvalConstr lang) => SymEval lang a -> SymEval lang a
 prune xs = SymEval $
-  StateT $ \st -> do
-    (x, st') <- runSymEvalRaw st xs
-    -- trace ("  X is " ++ show (pretty x)) $ return ()
-    -- trace ("  in the context " ++ show (pretty st')) $ return ()
-    solver <- ask
-    defs <- lift $ lift getPrtOrderedDefs
-    ok <- liftIO $ PureSMT.solveProblem (CheckPath $ CheckPathProblem st' defs) solver
-    guard ok
+  ReaderT $ \defs -> ReaderT $ \solvers -> StateT $ \st -> do
+    (x, st') <- runSymEvalRaw solvers defs st xs
+    guard (sesPath solvers (CheckPathProblem st' defs))
     return (x, st')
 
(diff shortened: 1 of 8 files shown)


VictorCMiraldo:
@0xd34df00d, a quick google gave me this: https://stackoverflow.com/questions/23471594/concurrent-stack-implementation-using-mvar/23499300#23499300

0xd34df00d:
> @0xd34df00d, a quick google gave me this: https://stackoverflow.com/questions/23471594/concurrent-stack-implementation-using-mvar/23499300#23499300

Oh welp, noticed this just now, sorry! Anyway, using `QSem` seems cleaner than what's in that answer.

serras:
Did GitHub get confused by the two PRs?

VictorCMiraldo:
Either GitHub, Git or I got confused by the two PRs; I rebased this branch not inclusing the two initial commits on the latest `dev` (after merging #109) but now the web interface shows more commits than it should

serras:
I think it's good now. I know that if you force-push GitHub leaves anything above it unchanged, but the Files tab work as it should.

src/PureSMT.hs
```diff
@@ -22,3 +29,59 @@ class Solve domain where
...
+launchAll :: forall domain . Solve domain => Ctx domain -> IO [MVar X.Solver]
+launchAll ctx = replicateM numCapabilities $ do
+  s <- X.launchSolverWithFinalizer "cvc4 --lang=smt2 --incremental --fmf-fun" debug0
+  initSolver @domain ctx s
+  newMVar s
+  where
+    -- TODO: these constants should become parameters at some point; the solver command too!
+
+    numCapabilities :: Int
+    numCapabilities = 3
```
serras, line 63: Why not start with the [RTS threading argument `numCapabilities`](https://hackage.haskell.org/package/base-4.14.1.0/docs/GHC-Conc.html#v:numCapabilities)?

src/Pirouette/Symbolic/Eval.hs
```diff
@@ -327,19 +352,30 @@ symEvalOneStep t@(R.App hd args) = case hd of
           R.TermArg
           args
 
+-- A-ha... it's the state monad annoying us. Maybe if we can understand whether
+-- we really need the state monad we can better understand the parallelism possibilities
+-- in here. If we do really need the monad, we can only paralellize the prune and pruneAndValidate
+-- functions, if not, we could paralellize everything.
```
serras, line 358: We are actually using `State` mostly to nest changes to the statistics. In fact, it should be possible to create a version of `ReaderT` which gives you back the "innermost" thing it used for its `local` or something like that.

serras (approved):

VictorCMiraldo:
For the record; it's remarkable that Pirouette is significantly simpler and faster! This is CI's time for the `isUnity` test:
```
      isUnity Bug
        [correct_isUnity v] validate [\r _ -> r] counter:                                                 OK (2.71s)
```
A few PRs ago we were close to 10s! :tada:

PR #111, opened 2022-06-17
A stab at mstacks

0xd34df00d:
The problem with the current implementation seems to stem from `MVar` both keeping the list of resources and signifying whether any resource is available. Looks like the only reasonable way is to keep track of the availability separately, for example, via an additional semaphore.

I have a feeling this shall work, and shall be deadlock- and race condition-free. We can discuss this separately.

### DIFF
--- a/src/PureSMT.hs
+++ b/src/PureSMT.hs
@@ -83,5 +83,5 @@ pushMStack a (sem, q) = do
 popMStack :: MStack a -> IO a
 popMStack (sem, q) = do
   waitQSem sem
-  modifyMVar q $ \case []     -> error "invariant disrespected; MStack should never be empty"
+  modifyMVar q $ \case []     -> error "invariant disrespected; MStack should not be empty if QSem gives passage"
                        (x:xs) -> pure (xs, x)
--- a/src/PureSMT.hs
+++ b/src/PureSMT.hs
@@ -2,11 +2,14 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE LambdaCase #-}
+
 module PureSMT (module X, Solve(..), solve) where
 
 import PureSMT.Process as X
 import PureSMT.SExpr as X
 import Control.Concurrent.MVar
+import Control.Concurrent.QSem
 import System.IO.Unsafe (unsafePerformIO)
 import Control.Monad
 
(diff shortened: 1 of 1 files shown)


src/PureSMT.hs
```diff
@@ -62,55 +65,23 @@ launchAll ctx = replicateM numCapabilities $ do
...
 popMStack :: MStack a -> IO a
-popMStack q = do
-  xss <- takeMVar q
-  case xss of
-    [] -> error "invariant disrespected; MStack should never be empty"
-    [x] -> return x
-    (x:xs) -> putMVar q xs >> return x
+popMStack (sem, q) = do
+  waitQSem sem
+  modifyMVar q $ \case []     -> error "invariant disrespected; MStack should never be empty"
```
VictorCMiraldo, line 86:
```suggestion
  modifyMVar q $ \case []     -> error "invariant disrespected; MStack should not be empty if QSem gives passage"
```

VictorCMiraldo (approved):
Looks awesome, much cleaner than the stack overflow comment I posted on the other PR!
I do think it is deadlock free and the only slight inaccuracy is the comment below.
 
Now, even if we have `n` threads block on the `waitQSem` from `popMStack`, only one of them will be woken if some other thread `t` calls `signalQSem`, meaning that `t` will have pushed an `a` into the `MVar`, so the woken thread will have something to use.

aspiwack (commented):
Just passing by. I'd like to point out that for this sort of applications, `STM` is much easier. So you should always consider `STM` as the default, and have a reason to use `MVar`s instead (a reason may be, for instance, the need to wrap actual `IO` (like writing to a file) in a lock).

In this case, it would look like this:
```haskell
type MStack = TVar [a]

newMStack xs = newTVarIO xs

pushMStack x q = atomically $ do
  xs <- readTVar q
  writeTVar $ x : xs
  
popMStack q = atomically $ do
  xs <- readTVar q
  case xs of
    [] -> retry
    x:xs' -> do
      writeTVar q
      return x
```

The `STM` is always dealock-free. You can also expose the `STM`-typed function (rather than just the `IO` functions) so that consumers can make their own transaction. It does lock you into using `STM`, though.

VictorCMiraldo:
Thank you @aspiwack! That looks cleaner indeed. The main problem with atomically is that we're using `pushMStack` and `popMStack` inside a `unsafePerformIO`, and the [docs](https://hackage.haskell.org/package/stm-2.5.0.2/docs/Control-Monad-STM.html#v:atomically) say we shouldn't. Because this is a very small problem and code that is private to that module, I'm OK sticking to raw `MVar`s. WDYT?

aspiwack:
It is, indeed, a good reason to not use the `STM` that I wasn't thinking of.

PR #113, opened 2022-06-20
More Plutus IR built-ins

serras:
This implements quite some Plutus IR primitives, leaving aside those which have to make with encoding and cryptography, which require some thinking about their representation. Some additional notes:
- This PR introduces a separation between a type and a term namespace. This was the only way I could resolve identifiers to their proper meanings, given that Plutus IR code (following Haskell) uses a lot of type/constructor punning.
- The `Data` type is represented by an additional datatype defined internally.
- Some commonalities between the `Example` and `PlutusIR` evaluators have been moved into a single `Pirouette.Symbolic.Eval.Helpers` module.

### DIFF
--- a/src/Language/Pirouette/PlutusIR/SMT.hs
+++ b/src/Language/Pirouette/PlutusIR/SMT.hs
@@ -207,7 +207,7 @@ trPIRFun op [x, y] =
     P.EqualsData -> Just $ PureSMT.eq x y
     -- constructors
     P.MkCons -> Just $ PureSMT.fun "Cons" [x, y]
-    P.MkPairData -> Just $ PureSMT.tuple [x, y]
+    P.MkPairData -> Just $ PureSMT.fun "Tuple2" [x, y]
     _ ->
       error $
         "Translate builtin to SMT: "
--- a/src/Language/Pirouette/PlutusIR/SMT.hs
+++ b/src/Language/Pirouette/PlutusIR/SMT.hs
@@ -2,6 +2,7 @@
 
 module Language.Pirouette.PlutusIR.SMT where
 
+import Data.Maybe (isJust)
 import Data.Text (Text)
 import Language.Pirouette.PlutusIR.Syntax
 import Pirouette.Monad
@@ -10,7 +11,6 @@ import Pirouette.SMT.Base
 import Pirouette.SMT.Constraints
 import Pirouette.Term.Syntax.Base
 import Pirouette.Term.Syntax.SystemF as SystemF
-import Pirouette.Transformations.Contextualize (contextualizeTermName)
 import qualified PlutusCore as P
 import qualified PureSMT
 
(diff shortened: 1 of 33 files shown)


src/Language/Pirouette/PlutusIR/SMT.hs
```diff
@@ -83,103 +96,296 @@ trPIRFun :: P.DefaultFun -> [PureSMT.SExpr] -> Maybe PureSMT.SExpr
...
+  branchesBuiltinTerm P.UnBData _ [d] =
+    continueWithDataMatch 
+      d (TyArg (builtin PIRTypeByteString))
+      errorTerm errorTerm errorTerm errorTerm (App (Bound (Ann "b") 0) []) []
+
+  branchesBuiltinTerm _rest _translator _args = 
+    pure Nothing
+
+
+continueWith ::
```
0xd34df00d, line 344: Might be helpful to put a comment what `continueWith*` family of functions does.

src/Language/Pirouette/PlutusIR/ToTerm.hs
```diff
@@ -35,6 +36,13 @@ import Data.Function (on)
 
 -- * Translating 'PIR.Type's and 'PIR.Term's
 
+-- [HACK: Translation of 'Bool']
+-- Many PIR files define their own 'Bool' type as an ADT
+-- with their own 'True' and 'False'. However, this breaks
+-- our translation, since this 'Bool' is *not* the built-in.
+-- So instead we transform any usage of a 'PIRType' with
+-- 'Bool' into the built-in one.
```
0xd34df00d, line 44: How do you reckon doing this as a separate `PIR` transformation? My reasoning for this is that this code is complex enough to split any special cases out to a separate pass/function/module/etc.

Sure, this can be done separately later.

0xd34df00d (approved):

src/Language/Pirouette/PlutusIR/SMT.hs
```diff
@@ -1,76 +1,89 @@
...
-  isStuckBuiltin = error "isStuckBuiltin (t :: TermMeta PlutusIR meta): not yet impl"
 
-instance LanguageSymEval PlutusIR where
-  branchesBuiltinTerm _tm _translator _args = pure Nothing
+  isStuckBuiltin e
+    | termIsConstant e = True
+    | Just _ <- termIsMeta e = True
+  isStuckBuiltin (App (Free (Builtin op)) args)
+    | op `elem` plutusIRBasicOps || op `elem` plutusIRBasicRels
+    = let args' = map (\(TermArg a) -> a) args
```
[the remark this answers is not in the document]
serras: It's guaranteed if the term is well-typed (since those operations take no type arguments). But you are right that including a safeguard won't hurt either...

src/Language/Pirouette/PlutusIR/SMT.hs
serras: Done

Issue #97, opened 2022-05-25
In `dev`: exit when counter-example is found


serras:
In a perfect world we would like to finish the search for a counter-example in Pirouette once one is found. However, due to the intricate coupling between the `IO` part and the breadth-first search, more work is being done that it should be right now.

VictorCMiraldo:
Given we now have a pure SMT interface, I'd be curious to know if we get to stop on the first counter-example now.

serras:
With #115 this should be the case: the `Data.Foldable.find` should stop once the first one is found.

PR #115, opened 2022-06-23
Introduce our own TreeT monad

serras:
This PR completely removes `WeightedList` in favor of our own `TreeT`. This data type has been engineering to completely fit our purposes:
- at each node we keep track of the state, so we can see how it changes,
- when we branch we can attach a tag, so we can later see how it's used (for example, in symbolic evaluation we tag what is the constructor we use).


src/Pirouette/Symbolic/Eval.hs
```diff
@@ -52,49 +53,59 @@ module Pirouette.Symbolic.Eval
...
     solvePropProblem :: CheckPropertyProblem lang -> PruneResult
   }
 
+data Tag lang = FromBuiltin (BuiltinTerms lang)
+              | FromConstructor Name
+type SymEvalResult lang a = Tree (Tag lang) (SymEvalSt lang) a
+
 -- | A 'SymEval' is equivalent to a function with type:
 --
 -- > SymEvalSt lang -> SMT.Solver -> m [(a, SymEvalSt lang)]
```
lucaspena, line 102: Is this comment outdated? Should the right hand side mention trees now?

serras:
I've rebased this PR on top of the latest changes in `dev`, so it merges cleanly.

lucaspena:
On my end, the O'Hearn Peano test successfully terminates in 60-75 seconds (I ran it a few times). So it seems the tree is making the test take much longer, but it looks like it is terminating at least.

src/Pirouette/Symbolic/Eval.hs
```diff
@@ -88,18 +98,20 @@ data SymEvalEnv lang = SymEvalEnv
...
--- > SymEvalSt lang -> SMT.Solver -> m [(a, SymEvalSt lang)]
+-- It keeps the environment and a connection to the solver in a Reader,
+-- other than that it's the 'TreeT' monad.
 newtype SymEval lang a = SymEval
   { symEval ::
-      ReaderT
-        (SymEvalEnv lang)
-        (StateT (SymEvalSt lang) WeightedList)
-        a
+      ReaderT (SymEvalEnv lang) (TreeT (Tag lang) (SymEvalSt lang) Identity) a
```
VictorCMiraldo, line 111: If we use `TreeT tag Identity`, why not define it as `Tree tag` and explicitely not make it a transformer? Adding any monad other than `Identiy` at the bottom of this stack would be heavily discouraged anyway.

src/TreeT.hs
```diff
@@ -0,0 +1,207 @@
...
+                             TreeT tag state Identity a -> state -> Maybe r #-}
+
+-- | Defines a class of monad which support pruning and branching.
+class Monad m => MonadTree tag m | m -> tag where
+  stop :: m a
+  branch :: [(tag, m a)] -> m a
+
+-- | The monad which defines actions which keep a 'state',
+-- and also may prune and branch.
+data TreeT tag state m a where
```
VictorCMiraldo, line 147: This is much more than just lifting `Tree` to a transformer. Also, one of the purposes of this exercise was to explicitely detach the notion of state from the symbolic engine. With the `GetT` and `PutT` in here, we just hid that notion of state deeper. 

I tend to think it might be simpler to just make the one function that needed state return it explicitely. I haven't looked at it in detail though.

serras: I think that actually the notion of state in inherent in the engine, in all the constraints and information we are gathering. Maybe instead of `GetT` and `PutT` are not the best choices, and maybe `AddConstraintT` is the right move here, in the same way that `BranchT` has become a constructor in `TreeT`... 🤔 What do you think?

src/Pirouette/Symbolic/Eval.hs
serras: That sounds good to me. My initial reaction wa "what if we ever need to go back to `IO`?" But I guess that's for the future...

VictorCMiraldo:
I just noticed the existence of the `gr/unstate` branch from @0xd34df00d. How does the work in these two branches relate? It's good to communicate to make sure we're not overlapping work.

VictorCMiraldo:
Great stuff @lucaspena, thank you! Besides having empirical evidence that the `treet` branch is doing additional work, it reinforces the idea that currently, in `dev` we are returning the first counterexample we find and looking no further! (#97)

VictorCMiraldo:
The work of #137 will supersede this; this PR is stale and we decided not to move forward with it.

PR #117, opened 2022-06-25
Add some more documentation

facundominguez:


src/PureSMT.hs
```diff
@@ -66,7 +79,12 @@ launchAll ctx = replicateM numCapabilities $ do
 -- |An 'MStack a' is an 'MVar' having a list of 'a's,
 -- which can be popped off the list and pushed back onto it.
 -- Popping off an 'MStack' that has no available 'a's just blocks until one becomes available.
-type MStack a = (QSem, MVar [a])
+type MStack a =
+    -- Invariants:
+    -- * The qsem has no more resources than the length of the list in the MVar
+    -- * For every element added the list in the MVar, eventually one resource
+    -- is added in the qsem.
```
0xd34df00d, line 86: I'm not sure about the word `eventually` here — it can bring unrelated associations from contexts like eventual consistency.

I'd use something like this (although this wording is also perhaps non-optimal):
```suggestion
    -- * For every element added the list in the MVar, one resource
    -- is then added in the qsem (although not atomically).
```

0xd34df00d (approved):
LGTM modulo wording in one comment.

Issue #122, opened 2022-06-30
Finalize the PlutusIR interface


VictorCMiraldo:
We managed to symbolically evaluate some PlutusIR programs, but the interface needs work. If we are to rely on pirouette during our audits, we need to be able to easily run it over a validator. I can imagine that something that looks like the following would work:

```haskell
pirouettePIR ::
  TypedValidator a ->
  AssumeProve PlutusIR ->
  Assertion
pirouettePIR val condition =
  let pir = Pl.getPir val
      env@(PrtUnorderedDefs decls main) = _transformAndTr pir
   in assertIncorrectnessLogic env (proveAny _stoppingCond) $ IncorrectnessLogicParams
         { ipTarget = main,
           ipTargetType = [pir| Bool |],
           ipCondition = condition
         }
```

There are a few discrepancies that should be investigated and taken care of, though. For example:
1. The `main` term in a `PrtUnorderedDefs` is never used and can't be referred to. Maybe we should change `PirouetteReadDefs` to exclute that main term. It would certainly make things cleaner
2. When writing the `condition`, we will likely need to rely on existing definitions or plut in a set of auxiliary definitions. Do we need a pirouette-specific plutus prelude? For instance, say the validator is the [auction validator](https://github.com/tweag/plutus-libs/blob/main/examples/src/Auction.hs#L310), we might want to say that a "bid" never loses that thread token. How do we write this in:
    ```haskell
    cond :: AssumeProve PlutusIR
    cond = ( [pir| \(res : Bool) (dat : Datum) (red : Redeemer) (ctx : ScriptContext) . res && isBid red |] 
                :==>: [pir| \res dat red ctx . exists out in continuingOuts ctx such that ... |])
    ```
As we can see, that will get really difficult really fast, we need to find a way to write these properties with ease.

PR #123, opened 2022-07-01
Start refining the interface for invoking pirouette on `PlutusIR` (#122)

VictorCMiraldo:
This PR starts stabbing at #122:

- [x] Removes an explicit mention of a main term in `PrtUnorderedDefs` and `PrtOrderedDefs`. This was a technicality from PlutusIR and shouldn't affect the rest of pirouette.
- [ ] ~Define a `pirouettePIR` for interacting with validators~

After too much pain with figuring out dependencies, it occurred to me that maybe everything `PlutusIR` specific should fall into `plutus-libs`! I'll do that on this PR and make sure to define this as `pirouette-2.0.0` so I have a commit to depend on for the cabal file in `plutus-libs`.

---

Turns out this is doing significantly more than just working on the interface. As I'm moving forward I'm fixing a number of bugs. Namelly:

- Removes [`prtDefOfAnyNamespace`](https://github.com/tweag/pirouette/pull/123/files#diff-f916a708ed6174a1233f4bd8eda2056f01c9fedd5dc9fe735b5aa28bb0909b90L77) as certain constructor might be called the same as their types with the different namespaces. Propagate changes by [taking namespaces into account for dependency computation](https://github.com/tweag/pirouette/pull/123/commits/4918a5ef9990e40b2f125852c9ee5e1190735444).
- Fix a defunctionalization bug: the engine assumed that term and type variables came from the same index pool, in reality, type variable index 0 *is not* term variable index 0: this is a proper term: `f @0 #0`, its a function `f` applied to the first type variable and the first term variable.
- Adds a `PureSMT.Options` and a `Pirouette.Symbolic.Eval.Types.Options` so we can later control options through tasty options, otherwise debugging is a massive pain.
- Adds a `LanguagePrelude` class with a `builtinPrelude` method, replacing the `builtinTypeDefs` from `LanguageBuiltins`.
- Adds a simple symbolic eval statistics result to `proveAny`
- Monomorphizes all terms, which makes the original and simpler defunctionalization engine work alright (at least for now)


src/Language/Pirouette/QuasiQuoter/Syntax.hs
```diff
@@ -11,8 +11,7 @@
 -- writing terms for testing pirouette a with a little less work.
 -- A lot of the syntactical choices have been made to make parsing
 -- as easy as possible. A program in this language is a set of
--- definitions containing one definition named @main@.
--- One example program would be:
+-- definitions. One example program would be:
```
0xd34df00d, line 14: Kudos for updating comments as well!

src/Language/Pirouette/QuasiQuoter/Syntax.hs
```diff
@@ -121,16 +120,11 @@ deriving instance
 
 -- * Parsers
 
--- | Parses a program, consisting in a map of declarations and one distinguished
---  function called main.
+-- | A program consists in a number of declarations
```
0xd34df00d, line 123:
```suggestion
-- | A program consists of a number of declarations
```
or
```suggestion
-- | A program is some set of declarations
```
(the latter also emphasizes unordered structure at this point)

0xd34df00d (approved):

src/Language/Pirouette/QuasiQuoter/Syntax.hs
VictorCMiraldo: I like the second more! Will update! Thanks

VictorCMiraldo:
I'm working with this PR in tandem with https://github.com/tweag/plutus-libs/pull/133, once I get all the pirouette-specific bugs out of the way, I'll merge. I'll make sure to update the description accordingly!

VictorCMiraldo:
Seems like we're hitting another defunctionalization bug; this is the definition of a lengthy `_Apply!!....` function:
```
(λ (ctx : Closure!!PIRTypeByteString_PIRTypeByteString_Bool)
         (ctx : List (Tuple2 ByteString Integer)) (e6902 : Tuple2 ByteString
                                                                  Integer)
         (xs6903 : List (Tuple2 ByteString Integer)) . Tuple2_match @ByteString
           @Integer
           1#e6902                                                                                               
           @(List (Tuple2 ByteString Integer))                                                          
           (λ (c6905 : ByteString) (ds6906 : Integer)                                                
            . Bool_match (fFoldableNil_cfoldMap!Bool!Tuple2<PIRTypeByteString!PIRTypeInteger> 
                -- HERE!! Note how the CConsMonoid!Bool constructor got a lambda, not a Closure!!Bool_Bool_Bool!
                (CConsMonoid!Bool (λ (eta9615 : Bool) (eta9617 : Bool) . AdditiveMonoid_match!Bool v9613
                                                                                                                     @Bool
                                                                                                                     (λ
                                                                                                                      (v9622 : Bool -> Bool -> Bool)
                                                                                                                      (v9624 : Bool)
                                                                                                                      . 1#v9622 3#eta9615
                                                                                                                          2#eta9617))
                        (AdditiveMonoid_match!Bool v9613
                            @Bool
                            (λ (v9627 : Bool -> Bool -> Bool) (v9629 : Bool)
                             . 0#v9629)))
                    (Closure!!Tuple2<PIRTypeByteString!PIRTypeInteger>_Bool_ctor_1 5#ctx
                        1#c6905)
                    4#ctx)
                @(List (Tuple2 ByteString Integer))
                2#xs6903
                (Cons @(Tuple2 ByteString Integer) 3#e6902 2#xs6903)))
```

src/Language/Pirouette/QuasiQuoter/Syntax.hs
```diff
@@ -114,23 +113,19 @@ data Expr lang
   | ExprLit (Constants lang)
   | ExprBase (BuiltinTerms lang)
   | ExprIf (Ty lang) (Expr lang) (Expr lang) (Expr lang)
+  | ExprBottom (Ty lang)
```
0xd34df00d, line 116: Ok, I was reviewing this more carefully — why do we need bottoms again?

src/Pirouette/SMT/FromTerm.hs
```diff
@@ -161,10 +161,13 @@ constructorFromPIR ::
...
+constructorFromPIR tyVars (name, constructorType) = handleError $ do
   -- Fields of product types must be named: we append ids to the constructor name
   let fieldNames = map (((toSmtName name ++ "_") ++) . show) [1 :: Int ..]
       (_, unwrapped) = Raw.tyUnwrapBinders (length tyVars) constructorType
       (args, _) = Raw.tyFunArgs unwrapped
   cstrs <- zip fieldNames <$> mapM translateType args
   return (toSmtName name, cstrs)
+  where
+    handleError :: TranslatorT m a -> TranslatorT m a
+    handleError m = m `catchError` (throwError . (("At " ++ show name ++ ":") ++))
```
0xd34df00d, line 173: Also pointing to the function where this happened (`constructorFromPIR`) might be useful in the future.

0xd34df00d: Also, is this effectively [`withError`](https://hackage.haskell.org/package/mtl-2.3/docs/Control-Monad-Error-Class.html#v:withError)?

src/Pirouette/Symbolic/Eval/SMT.hs
```diff
@@ -113,8 +113,10 @@ instance (LanguageSMT lang) => PureSMT.Solve lang where
   type Ctx lang = SolverSharedCtx lang
   type Problem lang = SolverProblem lang
   initSolver SolverSharedCtx {..} s = hackSolver s $ do
-    _ <- runExceptT $ declareDatatypes solverCtxDatatypes
-    mapM_ (uncurry declareRawFun) solverCtxUninterpretedFuncs
+    res <- runExceptT $ declareDatatypes solverCtxDatatypes
+    case res of
+      Left err -> error ("Couldn't declare datatypes: " ++ show err)
```
0xd34df00d, line 118: Hmm, isn't `err` a string here, so wouldn't `show err` make quotes and other things it escapes slightly less readable?

src/Pirouette/Symbolic/Prover.hs
```diff
@@ -186,7 +188,7 @@ worker resultVar bodyTerm assumeTerm proveTerm = do
       -- debugPrint (pretty proveTerm')
       -- debugPrint somethingWasEval
       -- check the fuel
-      noMoreFuel <- gets sestStoppingCondition >>= \s -> s <$> currentStatistics
+      noMoreFuel <- asks (stoppingCondition . seeOptions) >>= \s -> s <$> currentStatistics
```
0xd34df00d, line 191: Probably worth taking a chance to rename the binding to smth like `shouldStop`, since it's not necessarily about the fuel now.

Also, the comment above is stale, I'd delete it because the code here is fairly straightforward.

0xd34df00d (approved):

src/Language/Pirouette/QuasiQuoter/Syntax.hs
VictorCMiraldo: Because otherwise we can't really write terms such as `fromJust` in a predicate, its definition would be:
```
fun fromJust : all (a : Type) . Maybe a -> a
  = /\ (a : Type) . \ (ma : Maybe a) . match_Maybe @a ma (\(x : a) -> a) (bottom @a)
```

This is useful because we might want to check the following predicate:
```
[term| (\(res : Result) (i : Maybe Integer) . res && isJust i) |] :==>: [term| (\res : Result) (i : Maybe Integer) . fromJust i > 10 |]
```

i.e., we know a property about the input in the antecedent of the implication, and we want to extract its value in the
consequent, but for that we need `fromJust`, which doesn't have a definition without `bottom`.

package.yaml
```diff
@@ -2,48 +2,38 @@ verbatim:
...
   - tasty-expected-failure
+  - QuickCheck
   - haskell-stack-trace-plugin
+  - parallel
   - optics-core
   - optics-th
-  - ansi-terminal
-  - parallel
+  # PureSMT deps
+  - process
```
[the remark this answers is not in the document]
VictorCMiraldo: To get the process id of the underlying solver. That's not something that `typed-process` re-exports:

```
import qualified System.Process as P
solverPid :: Solver -> IO (Maybe P.Pid)
solverPid = P.getPid . unsafeProcessHandle . process
```

We use the pid for debugging purposes only and it's really only useful when we have multiple servers running.

src/Pirouette/SMT/FromTerm.hs
VictorCMiraldo: We get the info that this is from `constructorFromPIR` from the `HasCallStack`. Indeed, this is `withError`, will change!

VictorCMiraldo: hmm... we're on an older `mtl` and it doesn't have `withError` easily reachable.

VictorCMiraldo:
Bingo! Managed to reproduce the issue with a better regression test for defunctionalization!

VictorCMiraldo:
Fixed one bug, found another!

Defunctionalization of destructors should also change the various type arguments involved; currently we are defunctionalizing the body of the destructors alright, but we fail to rewrite the type of values inside 'Just'. We're getting:

```
Maybe_match6620 @(ByteString -> BidderInfo7795) whatever @(Maybe6617 BidderInfo7795)
  -- Just case
  (λ ipv12631 : Closure!!PIRTypeByteString_BidderInfo . whatever)
  -- Nothing case
  whatever
```

But we should get:

```
Maybe_match6620 @Closure!!PIRTypeByteString_BidderInfo whatever @(Maybe6617 BidderInfo7795)
  -- Just case
  (λ ipv12631 : Closure!!PIRTypeByteString_BidderInfo . whatever)
  -- Nothing case
  whatever
```

Would be great to have a regression test for that.

PR #124, opened 2022-07-03
Monomoprhization and namespaces interaction bug

VictorCMiraldo:
The introduction of namespaces left out [one small detail on an auxiliary function](https://github.com/tweag/pirouette/commit/892da4fec6a932ba40a911a9c23bd4e7f5469af3#diff-88570a062606267f7e9d939af6469964c269bbd461b8a317cb47e79e2ad8269fR56), where it uses `TypeNamespace` for the names of the destructor and constructors, whereas it should use `TermNamespace` for those.


0xd34df00d (approved):
Good catch!

PR #129, opened 2022-07-09
Monomorphize Everything

VictorCMiraldo:
Simplifies the monomorphizer, which should help the defunctionalizer later by not having to deal with those nasty `Just @(Bool -> Bool)` cases: those will have been replaced by `Just!BoolToBool` and will be of type:

```Just!BoolToBool : (Bool -> Bool) -> Maybe (Bool -> Bool)```

Hence, it's obviously a higher-order constructor for the defunctionalizer!

Still incomplete: understand and decide how to deal with constructors of types that should be defunctionalized.

### DIFF
--- a/pirouette.cabal
+++ b/pirouette.cabal
@@ -54,6 +54,7 @@ library
       Pirouette.Transformations.Prenex
       Pirouette.Transformations.Term
       Pirouette.Transformations.Utils
+      Pirouette.Utils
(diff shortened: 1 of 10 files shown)


VictorCMiraldo:
I'll merge this on my branch then I'll simplify defunctionalization, with a little luck we solve everything!

Issue #132, opened 2022-07-13
Make `Eq` instance for `Datatype` into propositional equality


VictorCMiraldo:
https://github.com/tweag/pirouette/commit/5c809a2125b431714b01978884fecabe9b9738e2#r78340828

0xd34df00d:
Done.

PR #133, opened 2022-07-14
Add a separate class for α equality

0xd34df00d:
For #132.


src/Pirouette/Term/Syntax/Base.hs
```diff
@@ -290,6 +299,13 @@ data Definition lang
   | DTypeDef (TypeDef lang)
   deriving (Eq, Ord, Show, Data)
 
+instance (LanguageBuiltins lang) => AlphaEq (Definition lang) where
+  DFunDef f1 ~==~ DFunDef f2 = f1 == f2 -- TODO shall we do this modulo α-equivalence too?
```
VictorCMiraldo, line 303: Function definitions and types `Eq` instance is alpha-equivalence because we're using De Bruijn indices and `Ann`'s are ignored.

VictorCMiraldo (approved):

src/Pirouette/Term/Syntax/Base.hs
0xd34df00d: Sorry, that's a vague comment — I was thinking about doing the same for the `Ann` type that's used in terms (and thus in function definitions), and that right now is defined to relate any two elements of a particular type.

Probably not worth it though. LMK if you agree, and I'll remove the `TODO` and merge.

VictorCMiraldo: I wouldn't change the definition of equality for types and terms; the whole purpose of `Ann` is precisely to be just an annotation and not interfere with anything. For types and terms, `Eq` and `Ord` are compatible because they're both derived. 

I'm happy with removing the comment and merging :)

src/Pirouette/Term/Syntax/Base.hs
```diff
@@ -276,6 +299,13 @@ data Definition lang
...
```
VictorCMiraldo, line 303:
```suggestion
  -- the 'FunDef's default equality is already alpha-eq
  DFunDef f1 ~==~ DFunDef f2 = f1 == f2
```

src/Pirouette/Term/Syntax/Base.hs
VictorCMiraldo: Maybe something like the suggestion below would be nice, so we don't forget

PR #135, opened 2022-07-18
Improve the surface language syntax

### LINKED ISSUE #131: Improve the surface language syntax
Pirouette's `QuasiQuoter` parses a very baroque syntax, making it difficult to write more complex terms. That's important since we might need to augment a `PrtUnorderedDecls` with arbitrary definitions to support the predicates we want to run the prover with.

Having a syntax similar to Haskell could be valuable:
```
myHead :: forall a . List a -> Maybe a
myHead @a (Cons x xs) = Just @a x
myHead @a Nil = Nothing @a
```
There is a lot of freedom for the actual syntax and the above is just a suggestion:
- Maybe we can try threading type-variables automatically, when they're omitted...
- Maybe we start with simple case-statements instead of native pattern-matching...

Note that we should maintain the possibility of writing type-annotated lambdas, otherwise it will be too cumbersome to write the predicates as exemplified [here](https://github.com/tweag/plutus-libs/blob/bfd2b6264d64d470d2587a7f27696d6c5bab4bd0/examples/tests/AuctionSpec.hs#L271)

Also note that we should always be updating the pretty printer as mentioned in #126

florentc:
This introduces a new syntax to declare functions and fixes #131. It replaces the previous syntax. All the examples and tests have been rewritten accordingly. The dedicated test suite includes generic cases and regression tests for issues encountered during development.

# Features

- Haskell-like function declaration syntax
- Basic line-folding for top-level declarations (functions and datatypes) relying on a `Reader (Maybe Pos)` where `Pos` is the column where the current line-folding section (if any) began
- Type `@x` and term `x` parameters (no need for lambdas and LAMBDAS anymore although they are still supported)
- Case statements (syntaxic support only, translation to system F to be implemented) on patterns made of constructors, variables, and catch-all patterns `_`)
- Explicit or implicit kind for type variables in both function and datatype declarations (e.g. `forall (m : * -> *) a` is the same as `forall (m : * -> *) (a : *)`, and `data Maybe a` the same as `data Maybe (a : *)`)

```
isJust : forall a . Maybe a -> Bool
isJust @x mx =
  case @(Maybe x) @Bool mx of {
    Nothing -> False ;
    Just _ -> True
  }
```

# Other changes

- Migration of all example code and test suites to the new syntax
- Removal of the old syntax
- Extensive test suite for the new syntax
- Improved documentation in the parser module

# Future work

More detailed issues will be open after this PR is merged.

- Translation of case statements to system F. Once this is done, replace calls to `match_XXX` functions by case statements in examples (especially in `IsUnity`) and tests throughout Pirouette

- Nested line-folding: support for nested line-folded sections is not trivial but would make it possible to get rid of the `{ ; ; ; }` syntax in case statements and introduce `let` and `where` local function declarations (name clashes and name shadowing should be kept in mind when implementing this)

- Implicitly quantified type variables: `isJust : Maybe a -> Bool` would be equivalent to `isJust : forall a . Maybe a -> Bool`
- Implicit naming of type variables in function body: for `isJust : forall a . Maybe a -> Bool`, the body declaration `isJust mx =` would be equivalent to `isJust @a mx` (name clashes and name shadowing should be kept in mind when implementing this)

- Pattern matching directly in multiple function bodies: the `Param` type should rely on `Pattern lang` instead of `String`. The main question remains how to represent this when there are several pattern-matched function parameters: either as desugared nested case expressions `ExprCase` or in an adapted `FunDecl` type. The parser is ready for such a change and multi-bodies function declarations thanks to line folding, `Pattern lang`, and `Param`

- Adapt the pretty printer to provide output in the new syntax than can be parsed back


src/Language/Pirouette/QuasiQuoter/Syntax.hs
```diff
@@ -138,42 +169,104 @@ parseDecl =
...
+parseFunDecl = P.label "Function declaration (new syntax)" $ do
+  (r, funIdent, funType) <- lineFold $ do
+    r <- NonRec <$ symbol "nonrec" <|> pure Rec
+    funIdent <- ident @lang
+    parseTyOf
+    funType <- parseType
+    return (r, funIdent, funType)
+  lineFold $ do
+    _ <- symbol funIdent
+    -- TODO Fail when there are duplicate names in the term or type worlds
```
VictorCMiraldo, line 204: Is this TODO still relevant?

src/Language/Pirouette/QuasiQuoter/Syntax.hs
```diff
@@ -14,40 +15,52 @@
 -- definitions. One example program would be:
 --
 -- > -- Declares a datatype with a destructor named match_Maybe
--- > data Maybe (a : Type)
+-- > data Maybe (a : *)
```
VictorCMiraldo, line 18: Any reason for changing `Type` to `*`? I tend to prefer `Type` since we already have quite a few symbols. I thing that something like: `/\ (a : (Type -> Type) -> Type) . M` is more readable than `/\ (a : (* -> *) -> *)) . M`

src/Language/Pirouette/Example/IsUnity.hs
```diff
@@ -36,78 +36,89 @@ checkOk =
...
+lkupOne : forall k . forall v . (k -> Bool) -> Pair k v -> Maybe v
+lkupOne @k @v predi m =
+  match_Pair @k @v m @(Maybe v)
+    (\(fst : k) (snd : v) . if @(Maybe v) predi fst then Just @v snd else Nothing @v)
 
-fun lkup : all (k : Type)(v : Type) . (k -> k -> Bool) -> KVMap k v -> k -> Maybe v
-  = /\(k : Type)(v : Type) . \(eq : k -> k -> Bool)(m : KVMap k v) (tgt : k)
-  . match_KVMap @k @v m @(Maybe v)
-      (foldr @(Pair k v) @(Maybe v) (\(pk : Pair k v) . maybeSum @v (lkupOne @k @v (eq tgt) pk)) (Nothing @v))
+lkup : forall k . forall v . (k -> k -> Bool) -> KVMap k v -> k -> Maybe v
```
VictorCMiraldo, line 118: Can't we write `forall k v . (k -> k -> Bool) ... `?

src/Language/Pirouette/QuasiQuoter/Syntax.hs
```diff
@@ -138,42 +169,104 @@ parseDecl =
...
+  ExprAbs tp k (funTerm (substTyVarType i tp t2) ps body)
+funTerm (TyFun t1 t2) (Ident p : ps) body =
+  ExprLam p t1 (funTerm t2 ps body)
+funTerm _ [] body = body
+funTerm _ _ _ = error "Unexpected parameters in function declaration"
+
+-- | Apply a type variable name substitution taking into account naming
+-- collisions by renaming the conficting name beforehand.
+-- E.g. In `/\ a : * . /\ b : * . a -> b -> b`, by applying the type to `b`,
+-- the result will be: `/\ b_ : * . b -> b_ -> b_`
```
VictorCMiraldo, line 230:
```suggestion
-- E.g. In @/\ a : * . /\ b : * . a -> b -> b@, by applying the type to @b@,
-- the result will be: @/\ b_ : * . b -> b_ -> b_@
```

src/Language/Pirouette/QuasiQuoter/Syntax.hs
```diff
@@ -138,42 +169,104 @@ parseDecl =
...
 
-parens :: Parser a -> Parser a
-parens a = try (symbol "(") *> a <* symbol ")"
-
+-- | Parse a type. The following features are supported:
+--
+-- - Application: `Either a b`
+-- - Function arrow: `b -> (a -> b) -> c`
+-- - Quantification: `forall (a : * -> *) (b : *) . `
+-- - Type lambdas: `\(a : * -> *) (b : *) . `
```
VictorCMiraldo, line 267:
```suggestion
-- - Application: @Either a b@
-- - Function arrow: @b -> (a -> b) -> c@
-- - Quantification: @forall (a : * -> *) (b : *) . @
-- - Type lambdas: @\(a : * -> *) (b : *) . @
```

src/Language/Pirouette/QuasiQuoter/Syntax.hs
```diff
@@ -182,56 +275,80 @@ parseType = label "Type" $ makeExprParser pAtom [[InfixL pApp], [InfixR pFun]]
...
 
+-- | Parse an expression. The following features are supported:
+--
+-- - Type abstraction: `/\ (a : * -> *) (b : *) . `
+-- - Lambdas: `\ (x : Maybe a) (y : a) . `
+-- - If/then/else: `if @resultType (x == y) then expr1 else expr2`
+-- - Case statements:
+--   `case @type @resultType x of {pattern1 -> expr1 ; pattern2 -> expr2}`
+-- - Bottom: `bottom @resultType`
+-- - Type and term application: `f @a @b x @c y z`
```
VictorCMiraldo, line 325: Code blocks in haddock are written using `@`: https://haskell-haddock.readthedocs.io/en/latest/markup.html#code-blocks

```suggestion
-- - Type abstraction: @/\ (a : * -> *) (b : *) .@
-- - Lambdas: @\ (x : Maybe a) (y : a) . @
-- - If/then/else: @if \@resultType (x == y) then expr1 else expr2@
-- - Case statements:
--   @case \@type \@resultType x of {pattern1 -> expr1 ; pattern2 -> expr2}@
-- - Bottom: @bottom \@resultType@
-- - Type and term application: @f \@a \@b x \@c y z@
```

src/Language/Pirouette/QuasiQuoter/Syntax.hs
```diff
@@ -241,24 +358,52 @@ parseTerm = label "Term" $ makeExprParser pAtom ops
...
-      try (symbol "bottom")
+      P.try (symbol "bottom")
       ty <- symbol "@" >> parseTypeAtom
       return (ExprBottom ty)
 
+    parseCase :: Parser (Expr lang)
+    parseCase = do
+      P.try (symbol "case")
+      ty <- symbol "@" >> parseTypeAtom
+      tyRes <- symbol "@" >> parseTypeAtom
```
VictorCMiraldo, line 369: One particular thing that might make things a little easier is to shuffle some types around:
```
case @Bool xs @(List Int)
  { Nil -> False
  ; Cons _ _ -> True
  }
```

Now it's pretty evident what is the result type and there's hopefully less confusion since this would be congruent with if-statements

src/Language/Pirouette/QuasiQuoter/Syntax.hs
```diff
@@ -276,35 +421,133 @@ parseBinder binder parseVars parseBody = do
   b <- parseBody
   return $ foldr (uncurry binder) b vars
 
+-- | Parse a declaration of a type variable and its kind. The kind is `*` by
+-- default if omitted.
+-- E.g. `(a : * -> * -> *)`
+-- E.g. `(a : *)` and `a` are equivalent
+parseTypeVarKind :: forall lang. LanguageParser lang => Parser (String, SystF.Kind)
+parseTypeVarKind = parens p <|> p
```
VictorCMiraldo, line 429: Is the `parens` really necessary here? Shouldn't it be the responsibility of the caller to add `parens parseTypeVarKind` in the places it needs it?

src/Language/Pirouette/QuasiQuoter/Syntax.hs
```diff
@@ -276,35 +421,133 @@ parseBinder binder parseVars parseBody = do
...
+      "bottom"
+    ]
+
+-- | Parse an expression in parenthesis
+parens :: Parser a -> Parser a
+parens a = P.try (symbol "(") *> a <* symbol ")"
+
+-- | Parse an expression in curly brackets
+curlyBrackets :: Parser a -> Parser a
+curlyBrackets a = P.try (symbol "{") *> a <* symbol "}"
```
VictorCMiraldo, line 517:
```suggestion
braces :: Parser a -> Parser a
braces a = P.try (symbol "{") *> a <* symbol "}"
```

tests/unit/Language/Pirouette/ExampleSpec.hs
```diff
@@ -17,30 +21,208 @@ canParseType _ = return ()
...
     testCase "Can parse program" $
       canParseProgram
         [prog|
-          data Either (a : Type) (b : Type) = Left : Either a b | Right : Either a b
+          data Either (a : *) (b : *) = Left : Either a b | Right : Either a b
+          main : Integer
+          main = 42
+        |],
+    testGroup
+      "New syntax for function declaration"
```
VictorCMiraldo, line 56: While it is nice to have tests, we should watch out for too many tests. The test tree above is basically subsumed by this much more involved tests. Theonly one that might be worth keeping is the `"Type binders with or without parenthesis"` tests, which is a regression test from a previous bug on that.

tests/unit/Language/Pirouette/ExampleSpec.hs
```diff
@@ -17,30 +21,208 @@ canParseType _ = return ()
...
-          canParseTerm [term| /\ (f : Type -> Type) (x : Type) . \k : f (Either x Bool) . fmap whatever k |]
+          canParseTerm [term| /\ (f : * -> *) (x : *) . \k : f (Either x Bool) . fmap whatever k |]
       ],
     testCase "Can parse program" $
       canParseProgram
         [prog|
-          data Either (a : Type) (b : Type) = Left : Either a b | Right : Either a b
+          data Either (a : *) (b : *) = Left : Either a b | Right : Either a b
+          main : Integer
+          main = 42
```
VictorCMiraldo, line 53: There's no need to have a main term anymore :)

tests/unit/Pirouette/Term/Syntax/SystemFSpec.hs
```diff
@@ -23,14 +23,14 @@ tests :: [TestTree]
...
       sameTy
-        [ty| \f : Type -> Type . (\x : Type . f x) (\w : Type . w) |]
-        [ty| \f : Type -> Type . f (\w : Type . w) |],
+        [ty| \f : * -> * . (\x : * . f x) (\w : * . w) |]
+        [ty| \f : * -> * . f (\w : * . w) |],
     testCase "Type-Type application, #2" $
       sameTy
-        [ty| \y : Type -> Type . \z : Type -> Type . (\x : Type . y x) (\w : Type . z w) |]
-        [ty| \y : Type -> Type . \z : Type -> Type . y (\w : Type . z w) |],
+        [ty| \y : * -> * . \z : * -> * . (\x : * . y x) (\w : * . z w) |]
```
VictorCMiraldo, line 30: This one is hard to read with `*` instead of `Type`, for example.

VictorCMiraldo (commented):
This is starting to look pretty nice, thanks @florentc! There are still a couple loose ends that might need some final touches

src/Language/Pirouette/QuasiQuoter/Syntax.hs
florentc: It was. Now implemented :+1:

src/Language/Pirouette/Example/IsUnity.hs
florentc: We can. Some of these had evaded my refactoring pass. I also changed some `data Something (a: *) (b : *)` into `data Something a b` that were still left in their desugared form. This is now done.

src/Language/Pirouette/QuasiQuoter/Syntax.hs
florentc: It was used in all call sites of `parseTypeVarKind` so I moved it in the parser definition but I conceptually agree and moved back `parens` to call sites.

tests/unit/Language/Pirouette/ExampleSpec.hs
florentc: Indeed. Thank you. Done.

src/Language/Pirouette/QuasiQuoter/Syntax.hs
florentc: The main reasons were:
- Being closer to haskell syntax and what haskellers are used to, since the general syntax aims at this
- I actually think it makes the code more readable and concise. I think the `*` makes it clear immediately we are on typevar/kind level whereas the `a : Type` looks too similar to the equivalent at termvar/type level. I feel it requires slightly more brain power to disambiguate when skimming through code and sections with nested small and big lambdas. Also it takes up less space and is more convenient to write. This is absolutely not a hill I am willing to die on though, I am fine with both :smiley:

tests/unit/Pirouette/Term/Syntax/SystemFSpec.hs
florentc: I think it is mostly due to lack of parenthesis and also we can omit the kind when it is `*`/`Type`. This can look like `[ty| \y : (* -> *) . \z : (* -> *) . (\x . y x) (\w . z w) |]`

Again, not a strong opinion for/against `*`/`Type`. I am ok to change back if this really bothers you.

tests/unit/Language/Pirouette/ExampleSpec.hs
florentc: I agree the test tree should be pruned. However I think there are quite some other test cases that deserve to remain. They correspond to regression tests from issues during development of the parser. For example line folding.

Could we maybe lower the depth of the test tree and have only one combined test case for line folding, one for a complex case statement showcasing everything, one for function definition whose type has nested foralls with many variables and name shadowing?

src/Language/Pirouette/QuasiQuoter/Syntax.hs
florentc: Done :+1:

src/Language/Pirouette/QuasiQuoter/Syntax.hs
florentc: Are you ok to put the second `@type` after the `of` keyword? I think of possible parsing issues: the value we match against patterns could end with a type application and it may be hard to read and parse.

src/Language/Pirouette/QuasiQuoter/Syntax.hs
florentc: Rolled back to `Type` :+1:

tests/unit/Language/Pirouette/ExampleSpec.hs
florentc: Test suite has been pruned a lot.

tests/unit/Pirouette/Term/Syntax/SystemFSpec.hs
florentc: Rolled back to `Type`

VictorCMiraldo (approved):
This looks pretty nice, thank you @florentc!

VictorCMiraldo (approved):

PR #136, opened 2022-07-19
Use ghc 9.0.2: fixing nixpkgs HLS

VictorCMiraldo:
A few commits back we started relying on `haskell-language-server` from `nixpkgs` instead of compiling our own. That lead to issues because it was compiled with ghc 9.0.2 and we were building pirouette with ghc 8.10.7; This PR makes pirouette build with ghc 9.0.2;

### DIFF
--- a/nix/packages.nix
+++ b/nix/packages.nix
@@ -1,8 +1,5 @@
 { 
   sources ? import ./sources.nix {},
-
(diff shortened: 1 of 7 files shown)


florentc (approved):
Build and HLS working correctly :+1:

Issue #138, opened 2022-07-21
Remove the need for `StateT` in the `SymEval` monad


VictorCMiraldo:
Related to #137; we only have s `StateT` in our `SymEval` monad because of two reasons:

1. To generate unique names, but this is much more elegantly achieved through an approach such as ["On Generating Unique Names"](https://www.cambridge.org/core/services/aop-cambridge-core/content/view/763DE73EB4761FDF681A613BE0E98443/S0956796800000988a.pdf/functional_pearl_on_generating_unique_names.pdf), which is already implemented in the [value-supply](https://hackage.haskell.org/package/value-supply-0.6) library
2. To communicate constraints through the assume, prove and body terms in the prover. But this could be more easily handled by symbolically evaluating something like ` if assume (f i) i then (if prove i then () else bottom) else ()`, and make symbolic evaluation look for bottoms instead.

VictorCMiraldo:
One interesting option here might be to define a "language transformer"

```haskell
data SE lang

instance LanguageBuiltins lang => LanguageBuiltins (SE lang) where
  type BuiltinTerms = Either Triple (BuiltinTerms lang)
   ...
```

Now we can inject any builtins we might need to treat in a special way directly into the symbolic engine.

VictorCMiraldo:
My attempt at (1) above is here: https://gist.github.com/VictorCMiraldo/0a0847759c1702efa96270b2b397e650

Issue #137, opened 2022-07-19
How to deal with strict evaluation when arguments evaluate to bottom?


carlhammann:
Here's a problem @VictorCMiraldo and I discovered while discussing improvements to Priouette's symbolic evaluation.

Consider calling `f x y z = x == z`  with arguments like `f 42 (error "blah") 10`. Under strict evaluation (read: in Plutus), that function call would fail, but Pirouette currently inlines everything, which means that such a failure won't be detected.

In the incorrectness logic setting that means false positive counterexamples: Pirouette might think that the validator `f` returns`True` in situations in which it actually fails, even before it is called.

This issue is related to symbolic evaluation in the following way: We are thinking about symbolic execution into two phases, one function
```haskell
ana :: Term -> Tree a
```
that takes a term and builds a (potentially infinite) tree corresponding to the possible unfoldings of the given term, and a function
```haskell
cata :: Tree a -> [a]
```
to traverse that three and return the desired results. The tree data type should be essentially of the form
```haskell
data Tree a
  = Branch [Tree a]
  | Leaf a
  | Learn Constraint (Tree a)
   ...
```
so that we can write, for example (in pseudo-Haskell)
```haskell
ana [| case x of A -> u; B -> v |] = do
  res <- ana x
  Branch [ Learn (res := A) (ana u), Learn (res := b) (ana v) ]
```
The important observation here is that `ana` has made progress, but how do we make progress for something like `ana [| f x y z |]` without either enforcing lazy evaluation or descending in a potentially infinite recursion? Which additional constructor(s) of `Tree` should we have?

VictorCMiraldo:
As discussed earlier today with @carlhammann, we have a hunch that maybe the `Tree` type should capture the application structure, and it is up to the consumer to decide how to "combine" things. For instance, add the following constructor to `Tree a` (in pseudo-haskell)
```haskell
data Tree a = ...
   | Call (Vec n (Tree b) -> Tree a) (Vec n (Tree b))
```

When traversing this tree and with the intention of producing a list of results, we're met with two options:
```haskell
cataA :: Tree a -> [a]
cataA (Call f args) = cataA (f args)
```
or
```haskell
cataB :: Tree a -> [a]
cataB (Call f args) = cataB $ do
  args' <- distr args -- distr :: [Tree a] -> Tree [a]
  f $ map Leaf args'
```

[...]

PR #139, opened 2022-07-24
Rewrite the symbolic evaluation engine denotationally

VictorCMiraldo:
This PR aims to split the symbolic evaluation engine in its core parts:
1. One anamorphism that gives semantics to `Term lang` as an infinite tree of possibilities
2. Multiple catamorphisms that consume these trees according to some language parameters that will be provided by the user, things like:
    1. Do we run in call-by-name or call-by-value fashion?
    2. When to stop
    3. How to evaluate builtins
    4. etc...

This is in a very drafty state still, but for those interested, the discussion in #137 and #138 is what is driving the design.

### DIFF
--- a/pirouette.cabal
+++ b/pirouette.cabal
@@ -58,6 +58,7 @@ library
       PureSMT
       PureSMT.Process
       PureSMT.SExpr
+      Supply
   other-modules:
       Paths_pirouette
   autogen-modules:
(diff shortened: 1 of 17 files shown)


VictorCMiraldo:
As discussed with different people, this PR should be closed. The challenges surrounding the handling of builtins are still unclear. I do believe the juggling spectacle of writing the evaluator as a hylomorphism is possible, but it would require much more effort and time and initially predicted. I'm closing this PR as a result, please revive it for inspiration later, if needed!

PR #145, opened 2022-08-27
Remove expandAllNonRec

0xd34df00d:
It's not used elsewhere for quite some time, and it also has no tests/spec.

I was going through the transformations updating the comments, cleaning some things up, etc, and was curious whether we want to keep things we don't use anywhere, or whether we better remove them (git remembers everything after all!).

Making this as a separate PR, so it could be rejected easily in case you don't think it's useful!

### DIFF
--- a/src/Pirouette/Transformations/Inline.hs
+++ b/src/Pirouette/Transformations/Inline.hs
@@ -7,45 +7,10 @@
 module Pirouette.Transformations.Inline where
 
 import Data.Generics.Uniplate.Operations
-import Data.List (foldl')
 import qualified Data.Map as Map
-import qualified Data.Set as Set
-import Pirouette.Monad
 import Pirouette.Term.Syntax
 import qualified Pirouette.Term.Syntax.SystemF as SystF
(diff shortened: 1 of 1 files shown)


mmontin:
I think it's useful to remove whatever is no longer used.
There's already a lot to get into if one wants to get used to Pirouette so there's definitely value in trimming dead code. 
What was this piece of code used for before ?

VictorCMiraldo:
This makes sense! In fact, if pirouette were to become really language agnostic at some point, we can't have a `expandAllNonRec` that fixes a specific reduction semantics. Is it call-by-value, call-by-need or call-by-name? On that topic, I actually think we should refactor `elimEvenOddMutRec` in two parts: one which separates the mutually-recursive classes and another which expands terms.

To answer @mmontin, we had `expandAllNonRec` from back in the day when TLA+ was the target, it wasn't particularly important at any point tbh, given that it doesn't really make a difference and it has no longer been in use for a while now. Hence, +1 for removing it.

VictorCMiraldo (approved):

PR #146, opened 2022-08-29
Update Constraint Datatype

VictorCMiraldo:
This PR changes the simple `Constraints` datatype into `ConstraintSet`, adding a little more structure, namelly, it adds a `conjunct :: Constraint lang meta -> ConstraintSet lang meta -> Maybe (ConstraintSet lang meta)` which returns `Nothing` whenever it the resulting `ConstraintSet` would be trivially UNSAT.

-------------

edit: This PR is ready for review, someone should really tackle the implementation of a proper union sort, as suggested [here](https://github.com/tweag/pirouette/pull/146/files#diff-7e3619c59fa2491c5c175fc31b31cd4be309331bd044763b2dbab8e39953a376R101) (and write good tests for that!); It might be worth to do it as a separate issue, after merging.

------------

Unrelated important change: [this](https://github.com/tweag/pirouette/pull/146/files#diff-559aa0c21b712f0ebcbdc1eedea29956e48f7a1bd6e9815e0182f0af5f5c307dL216-R221) change in the order of terms that are executed by the `worker` function is very important! We should always first try to expand the assumption term, which might prune all sorts of uninteresting branches.


VictorCMiraldo:
The last commit introduces a hack for something we're not quite knowledgeable of: apparently, translation to SMT can fail and we would like to know exactly why and the implications of that.

VictorCMiraldo:
Added some debugging code in [here](https://github.com/tweag/pirouette/pull/146/files#diff-559aa0c21b712f0ebcbdc1eedea29956e48f7a1bd6e9815e0182f0af5f5c307dR162); plan for tomorrow is use the same debugging code on `main` and see how differently it behaves. I think we're pretty close.

VictorCMiraldo:
While debugging this morning, I quickly spotted the issue. On the `main` branch, we collect the following constraints for `cabal run spec -- -p /ohearn/ && /y == 11/`:
```
b/sub $s2 1 == 0
∧ $s2 /= 0
∧ x ↦ D $s2 $s3
∧ __result ↦ D $s0 $s1
Body:   b/ite @Delta False (D $s2 42) (D $s2 $s3)
Assume: b/== $s1 42
Prove:  b/== $s3 11
```

Note how we actively collect `b/sub $s2 1 == 0` and keep that! In our implementation of constraints, that is impossible to represent: we would gather that through `TermEq [term| b/sub $s2 1 |] [term| 0 |]`, but that will trigger the unification algorithm to run and that will say /nope/: those are not unifiable. Consequently, that branch would be automatically pruned out and we would miss the counterexample, making the test run forever!

We need to record non-unification constraints in our `ConstraintSet` too and engineer it a little more carefully. A great plan for today would be to fix the bug then write tests for that module!

src/Pirouette/SMT/Constraints.hs
```diff
@@ -1,167 +1,276 @@
...
+  ConstraintSet lang meta ->
   m (Bool, UsedAnyUFs, PureSMT.SExpr)
-constraintToSExpr knownNames (And constraints) = do
-  atomTrads <- mapM (runTranslator . atomicConstraintToSExpr knownNames) constraints
-  let (translations, usedUFs) = unzip (rights atomTrads)
-  return (all isRight atomTrads, mconcat usedUFs, PureSMT.andMany translations)
-constraintToSExpr _ Bot = return (True, NotUsedAnyUFs, PureSMT.bool False)
+constraintSetToSExpr knownNames ConstraintSet {..} = do
+  eAssignments <- mapM (runTranslator . uncurry trAssignment) $ M.toList csAssignments
+  eNotEq <- mapM (runTranslator . trRelations) csRelations
```
lucaspena, line 200: This is related to my other comment, but why is this called `eNotEq`? Equality also seems to be translated

src/Pirouette/SMT/Constraints.hs
```diff
@@ -1,167 +1,276 @@
...
+    -- Productive in the sense that we will always have at least one constructor
+    -- on the rhs.
+    csAssignments :: M.Map meta (TermMeta lang meta),
+    -- | Represents the symbolic variable equivalences we know of. Satisfies the invariant
+    -- that if @M.lookup v csmetaEq == Just v'@, then @v' < v@ (lexicographically)
+    csMetaEq :: M.Map meta meta,
+    -- | Any potential native constraint we might need.
+    csNative :: [PureSMT.SExpr],
+    -- | And finally, pairs of terms we've discovered /not/ to be equal
+    csRelations :: [(TermRelation, TermMeta lang meta, TermMeta lang meta)]
```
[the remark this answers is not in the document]
VictorCMiraldo: You're right, the comment is outdated. IIRC, we first had something like:
```
csNotEq :: [(TermMeta lang meta, TermMeta lang meta)]
```

src/Pirouette/SMT/Constraints.hs
```diff
@@ -1,167 +1,276 @@
...
-  = And [AtomicConstraint lang meta]
-  | Bot
-  deriving (Eq, Show)
+instance (Ord meta) => Semigroup (ConstraintSet lang meta) where
+  s1 <> s2 =
+    ConstraintSet
+      (csAssignments s1 <> csAssignments s2)
+      (csMetaEq s1 <> csMetaEq s2)
+      (csNative s1 <> csNative s2)
+      (csRelations s1 <> csRelations s2)
```
Niols, line 49: This will take the left-biased union of assignments and equalities, erasing keys of `s2` that are already present in `s1`. Isn't that a bit dangerous? Or does it actually not pose any problem ever? Also, I am not convinced that it conserves the invariants of the map of equalities. (It might even lose some equalities?)

I'm in the process of replaccing `csAssignments` and `csMetaEq` by a union-find implementation, so that should take care of the latter. For the former, wouldn't it make more sense to require `Semigroup meta` and to use `unionWith (<>)`? Will we even have `Semigroup meta`?

src/Pirouette/SMT/Constraints.hs
```diff
@@ -1,167 +1,276 @@
...
-  | Native PureSMT.SExpr
-  deriving (Eq, Show)
+data ConstraintSet lang meta = ConstraintSet
+  { -- | Represents the set of productive symbolic variable assignments we know of.
+    -- Productive in the sense that we will always have at least one constructor
+    -- on the rhs.
+    csAssignments :: M.Map meta (TermMeta lang meta),
+    -- | Represents the symbolic variable equivalences we know of. Satisfies the invariant
+    -- that if @M.lookup v csmetaEq == Just v'@, then @v' < v@ (lexicographically)
+    csMetaEq :: M.Map meta meta,
```
Niols, line 36: Is it possible to know of an equivalence between two meta-variables without knowing any assignment of them? My first instinct was that, no, it wasn't possible. But now I'm pretty sure it could be. Whatever the answer is, we should probably specify it. Also, if it is possible, then some code later looks quite fishy to me.

Niols: In particular, the notion of [“new `v`”](https://github.com/tweag/pirouette/blob/f12b95104b8b17996228587867c85f193944b7ab/src/Pirouette/SMT/Constraints.hs#L146) in `unifyMetaWith` is a bit counter-intuitive, because it is not actually new, just without assignment. Using a union-find (which I'm in the process of doing) would make this part of the code clearer, but I need to be sure about what the semantics is *supposed to be* before making sure that I respect it.

qaristote:
When I merge this to `qa/z3_inline-c_faster-eval_batches` and run the benchmarks the running-time of both Z3 and Pirouette seem to increase by ~2s (out of 20s for Z3 and 40s for Pirouette).

qaristote:
More precisely, here's the kind of benchmark results I get (after the PR / before the PR):
```
          Z3: 22.984s  count:  113600
     command: 30.788s  count: 1546100
launchSolver:  0.412s  count:     100
        main: 42.496s  count:       1
   readSExpr:  0.026s  count:   56800
  showsSExpr:  7.390s  count: 1659700

          Z3: 21.842s  count:  113600
     command: 29.886s  count: 2355500
launchSolver:  0.432s  count:     100
        main: 43.256s  count:       1
   readSExpr:  0.028s  count:   56800
  showsSExpr:  7.442s  count: 2469100
```
The time difference varies but the pre-PR version is consistently faster than the post-PR one. I don't see how the post-PR version could be sending less commands to the solver but still spend more time inside it.

qaristote:
Something else I haven't mentioned here is that the `Z3` count is only incremented when a command with interesting results is sent, typically `check-sat` (whereas the `command` count is incremented for every command). So I would expect the `Z3` count to be smaller post-PR since it should not be checking for satisfiability on trivial problems.

> So, it makes intuitive sense to me that there are fewer problems the solver sees, but it has to think harder about them.
I agree with the fewer problems part, but why would the solver would have to think harder about them ? Aren't they supposed to be the same problems ?

>  Do you have any data on, say, the size of the problems?
Apart from the number of commands sent, no. What kind of data would help ?

qaristote:
I logged the commands sent to the solver pre- and post-PR and ran `diff` on the results. The output is huge but here are some interesting examples that I got by looking at the first lines (`<` is pre-PR and `>` is post-PR):
```
< (set-option :produce-models true)(declare-fun pir_greaterThan1 (Int) Bool)(declare-fun pir_greaterThan0 (Int) Bool)(declare-fun pir_add1 (Int) Int)(push 1)(declare-fun pir___result () Int)(declare-fun pir_x () Int)(check-sat)
---
> (set-option :produce-models true)(declare-fun pir_greaterThan1 (Int) Bool)(declare-fun pir_greaterThan0 (Int) Bool)(declare-fun pir_add1 (Int) Int)(push 1)(declare-fun pir___result () Int)(declare-fun pir_x () Int)(assert true)(check-sat)
```
and
```
< (pop 1)(push 1)(declare-fun pir___result () Bool)(declare-fun pir_s0 () Int)(declare-fun pir_x () pir_MaybeInt)(assert (= pir___result true))(assert (= pir_x ((as pir_JustInt pir_MaybeInt) pir_s0)))(check-sat)
---
> (pop 1)(push 1)(declare-fun pir___result () Bool)(declare-fun pir_s0 () Int)(declare-fun pir_x () pir_MaybeInt)(assert (and (= pir___result true) (= pir_x ((as pir_JustInt pir_MaybeInt) pir_s0))))(check-sat)
```
So it seems this PR is doing at least two things:
- when a problem is trivial, the solution is sent to the solver as an assert (c.f. the additional `assert true` in the first example)
- when several assertions are sent to the solver their are grouped under a conjunction (c.f. the second example).
The first point explains why the solver is sent as many problems pre- and post-PR, and the second point explains why there are still less commands sent to the solver overall (I assume the number of times Pirouette detects a problem is trivial is negligible compared to the number of assertions that can be merged). This still doesn't explain why the solver is spending more time on solving the problems post-PR though: if anything, it should be spending less time because some problems now come directly with their solutions.

qaristote:
I ran Z3 on the logfiles of the commands produced by Pirouette pre- and post-PR and I do get a noticeable difference in the running-times, although the files should be logically equivalent:

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `z3 isunity.pre >/dev/null` | 190.3 ± 12.6 | 176.4 | 231.9 | 1.00 |
| `z3 isunity.post >/dev/null` | 230.2 ± 3.0 | 223.6 | 241.0 | 1.21 ± 0.08 |

Looking at [the diff](https://gist.github.com/qaristote/1eac3c0e93828de6f15482ebf83c3da1) between both files more closely it really seems the only differences are that post-PR 1) a few trivial and negligible assertions are added, 2) some redundant declarations and assertions are removed and 3) assertions are merged together using conjunctions. This is to be taken with a grain of salt of course because there are too many lines for someone to go through every one of them.

My guess right now is that Z3 spends more time parsing one big assertion with conjunctions than it does parsing many small assertions consisting of simple (in)equalities. @facundominguez, you had a look at the internals of Z3's parser, does that sound plausible to you ?

facundominguez:
I have no intuition on the cost of parsing larger expressions. But it could be checked by splitting the conjunctions into separate assertions ahead of feeding them to the SMT solver.

facundominguez:
Ah, and there is a function that only does the parsing `Z3_parse_smtlib2_string`. You might be able to measure if it worsens with the large conjunctions. I modified the `oneshot.c` program to invoke it at some point:

```
#include<stdio.h>
#include<stdlib.h>
#include<stdarg.h>
#include<z3.h>

void my_parser_example(char const* const fileName) {
  Z3_config cfg;
  Z3_context ctx;
  cfg = Z3_mk_config();
  Z3_set_param_value(cfg, "model", "true");
  ctx = Z3_mk_context(cfg);
  Z3_del_config(cfg);

  FILE* file = fopen(fileName, "r");
  char* buffer = NULL;
  size_t len;
  ssize_t bytes_read = getdelim( &buffer, &len, '\0', file);

  Z3_ast_vector b = 
      Z3_parse_smtlib2_string(ctx, 
                              buffer,
                              0,
                                NULL,
                                NULL,
                                0,
                                NULL,
                                NULL);

  printf("%s", Z3_eval_smtlib2_string(ctx, buffer));
  fclose(file);
}

int main(int argc, char* argv[]) {
    char const* const fileName = argv[1];
    my_parser_example(fileName);
    return 0;
}
```

qaristote:
This doesn't seem to be the explanation: I merged the assertions on logfile produced by Pirouette pre-PR and I got

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `z3 isunity.default >/dev/null` | 200.6 ± 13.5 | 182.1 | 226.7 | 1.00 |
| `z3 isunity_merge.default >/dev/null` | 203.2 ± 3.6 | 197.2 | 227.0 | 1.01 ± 0.07 |

Niols:
Are those benchmarks still up-to-date with the merge of #150? I would expect so because #150 is not supposed to change the generated constraints, but who knows.

qaristote:
> Are those benchmarks still up-to-date with the merge of #150? I

Yep, I'm ran them again and I'm not seeing any difference.

Niols:
With `DEBUG_TIMESTATS_ENABLE=1 cabal run example`, I can confirm @qaristote's observations: this PR makes things slightly slower. However, the overall run is barely affected (1-2%). I suggest we merge this PR as it cleans up a lot of code for a very minor cost.

Niols (approved):

PR #147, opened 2022-08-29
Some documentation updates

0xd34df00d:
There's plenty to do on defunctionalization still, but I thought it's a good idea to get at least something merged first.


src/Pirouette/Transformations/ElimEvenOddMutRec.hs
```diff
@@ -16,8 +16,6 @@ import Pirouette.Term.TransitiveDeps
 import Pirouette.Transformations.Inline
 
 -- | Removes all Even-Odd mutually recursive functions from the program.
---  When successfull, sets the 'tord' state field with a list of names indicating the order in which
---  they should be defined so that the dependencies of a term @T@ are defined before @T@.
```
VictorCMiraldo, line 20: Why was this one removed? I believe this comment is still true

VictorCMiraldo (approved):
Always nice to see docs being updates! :tada:

src/Pirouette/Transformations/ElimEvenOddMutRec.hs
mmontin: @0xd34df00d let us know whether this comment needs to remain or not in the end, so we can merge this PR.

0xd34df00d: This seems to be the only mention of `tord` in the code base (at least, `git grep tord` thinks so). Moreover, there is no explicit state anymore, and I think, after the code has been updated, what this is saying is that the function returns `PrtOrderedDefs lang` (where there is an order-related field indeed). But it's already obvious from the type, so I thought about removing it.

I can add that back in the form of smth like

> When successful, returns a `PrtOrderedDefs lang`, which also contains the order in which the list of names should be defined.

if you think that's useful.

mmontin: To be honest, I don't know whether that's useful or not yet, I let you decide, but by default, I'm always up for adding relevant comment. Feel free to merge when you've decided to add it or not.

mmontin (approved):

PR #149, opened 2022-09-12
Type-level deps/revdeps for transformations

0xd34df00d:


src/Pirouette/Transformations/Monomorphization.hs
```diff
@@ -27,15 +27,22 @@ import Pirouette.Monad
...
 import qualified Pirouette.Term.Syntax.SystemF as SystF
+import Pirouette.Transformations.Prenex (Prenex)
+import Pirouette.Transformations.Tagged
 import Pirouette.Utils
 
 -- * Monomorphization
 
+data Monomorphized
+
+monomorphize :: (Language lang) => Xform '[Prenex] '[Monomorphized] (PrtUnorderedDefs lang)
```
mmontin, line 38: I'm wondering here if this should be
`Xform '[Prenex] '[Prenex, Monomorphized]`
In other words, is it implicit that transformations cannot "undo" any previous transformation ?

src/Pirouette/Transformations/Tagged.hs
```diff
@@ -0,0 +1,56 @@
...
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE TypeFamilyDependencies #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE ConstraintKinds #-}
+
+module Pirouette.Transformations.Tagged where
+
+import Data.Kind
+import GHC.TypeLits
+
```
mmontin, line 15: Do we need the following toy types ?

mmontin (changes requested):
Since the changes are not too involved according to me and do not make things unreadable, I believe we can merged this PR once :
- the few comments are addressed
- ormolu successfuly validates

src/Pirouette/Transformations/Tagged.hs
```diff
@@ -0,0 +1,56 @@
...
+data Foo
+data Bar
+data Baz
+
```
mmontin, line 19: I believe we no longer need these toy types.

src/Pirouette/Transformations/Monomorphization.hs
0xd34df00d: Yep, that's the assumption — transformations don't undo anything that happened before.

I agree this is not ideal, but I'm not sure there's a good, generic and fault-proof way to keep track of this: even if a transformation doesn't require something, it might accidentally break invariants established by some other transformation earlier.

If we care about such scenarios and if we think it's worth it, then the best approach would be to just do all the required transformations _inside_ the current one instead of keeping track of this "globally".

src/Pirouette/Transformations/Tagged.hs
0xd34df00d: Oops, I had a feeling I forgot to remove something, and here it is!

src/Pirouette/Transformations/Monomorphization.hs
Niols: For testing purposes, we could write dynamic checks for PNF / monomorphised code, etc. Basically a spec, which shouldn't be too hard to write. We could then check that transformations do indeed yield what they claim and check that other transformations do not violate the invariants. Of course, we'd only use that during testing and not when actually using the tool (although I suspect that such tests might be linear in the size of the inputs while symbolic exploration is exponential).

0xd34df00d:
As a side note, it also gives a fairly readable error messages:
```
src/Pirouette/Runner.hs:123:3: error:
    • Unsatisfied transformation pipeline dependencies: '[Prenex,
                                                          EtaExpanded]
    • In the first argument of ‘($)’, namely ‘MkPipeline’
```

PR #153, opened 2022-10-10
fix warnings

qaristote:
Building Pirouette produces a few warnings, namely:

- `-Wstar-is-type`
- `-Wunused-imports`
- `-Wunused-top-binds`
- `-Wunused-matches`
- `-Wname-shadowing`

This PR fixes the pieces of code producing these warnings.

### DIFF
--- a/src/Language/Pirouette/QuasiQuoter/ToTerm.hs
+++ b/src/Language/Pirouette/QuasiQuoter/ToTerm.hs
@@ -89,7 +89,7 @@ trTerm tyEnv termEnv (ExprIf ty c t e) = do
   e' <- trTerm tyEnv termEnv e
   return $ ifThenElse ty' c' t' e'
 -- trTerm tyEnv termEnv (ExprCase ty tyRes t cases) = trTerm tyEnv termEnv (ExprVar "undefined") -- TODO implement translation of case constructs
-trTerm tyEnv termEnv (ExprCase _tyRes _ty _t _cases) = throwError "translation of case expressions is not yet implemented"
+trTerm _ _ (ExprCase _ _ _ _) = throwError "translation of case expressions is not yet implemented"
 trTerm _ termEnv (ExprVar s) =
   case s `elemIndex` termEnv of
     Just i -> return $ SystF.termPure $ SystF.Bound (SystF.Ann $ fromString s) (fromIntegral i)
--- a/src/Language/Pirouette/QuasiQuoter/ToTerm.hs
+++ b/src/Language/Pirouette/QuasiQuoter/ToTerm.hs
@@ -89,7 +89,7 @@ trTerm tyEnv termEnv (ExprIf ty c t e) = do
   e' <- trTerm tyEnv termEnv e
   return $ ifThenElse ty' c' t' e'
 -- trTerm tyEnv termEnv (ExprCase ty tyRes t cases) = trTerm tyEnv termEnv (ExprVar "undefined") -- TODO implement translation of case constructs
-trTerm _ _ (ExprCase _ _ _ _) = throwError "translation of case expressions is not yet implemented"
+trTerm _tyEnv _termEnv (ExprCase _tyRes _ty _t _cases) = throwError "translation of case expressions is not yet implemented"
 trTerm _ termEnv (ExprVar s) =
   case s `elemIndex` termEnv of
     Just i -> return $ SystF.termPure $ SystF.Bound (SystF.Ann $ fromString s) (fromIntegral i)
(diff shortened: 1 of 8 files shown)


src/Language/Pirouette/QuasiQuoter/ToTerm.hs
```diff
@@ -89,7 +89,7 @@ trTerm tyEnv termEnv (ExprIf ty c t e) = do
   e' <- trTerm tyEnv termEnv e
   return $ ifThenElse ty' c' t' e'
 -- trTerm tyEnv termEnv (ExprCase ty tyRes t cases) = trTerm tyEnv termEnv (ExprVar "undefined") -- TODO implement translation of case constructs
-trTerm tyEnv termEnv (ExprCase _tyRes _ty _t _cases) = throwError "translation of case expressions is not yet implemented"
+trTerm _ _ (ExprCase _ _ _ _) = throwError "translation of case expressions is not yet implemented"
```
aspiwack, line 92: You can also silence this warning cheaper
```suggestion
trTerm _tyEnv _termEnv (ExprCase _tyRes _ty _t _cases) = throwError "translation of case expressions is not yet implemented"
```

src/Pirouette/SMT/Monadic.hs
```diff
@@ -117,10 +117,10 @@ declareDatatype typeName (Datatype _ typeVariables _ cstrs) = do
...
   return $ typeName : map fst cstrs
 
--- | Returns @Right ()@ when this datatype can be declared as a SMTLIB datatype
-supportedDatatype :: (LanguageSMT lang) => TypeDef lang -> Except String ()
-supportedDatatype (Datatype _ typeVariables _ cstrs) =
-  void $ runWriterT $ mapM (constructorFromPIR typeVariables) cstrs
+-- -- | Returns @Right ()@ when this datatype can be declared as a SMTLIB datatype
+-- supportedDatatype :: (LanguageSMT lang) => TypeDef lang -> Except String ()
+-- supportedDatatype (Datatype _ typeVariables _ cstrs) =
+--   void $ runWriterT $ mapM (constructorFromPIR typeVariables) cstrs
```
aspiwack, line 123: Plese don't leave dead code in. You can either export this function to mean that it's part of the interface, or delete it altogether (Git can always used to retrieve the function in the future)
```suggestion
```

src/Pirouette/Term/Syntax/Base.hs
```diff
@@ -371,9 +372,9 @@ type LanguageConstrs lang =
 -- no idea what that is, defining an empty instance will use the default
 -- implementatino of @builtinPrelude = M.empty@, which is a fine definition.
 class (LanguageConstrs lang) => LanguageBuiltins lang where
-  type BuiltinTypes lang :: *
-  type BuiltinTerms lang :: *
-  type Constants lang :: *
+  type BuiltinTypes lang :: Kind.Type
+  type BuiltinTerms lang :: Kind.Type
+  type Constants lang :: Kind.Type
```
aspiwack, line 377: `Type` is central enough to Haskell that it is customary to import it unqualified. I assume there is an internal type to Pirouette called `Type` as well for you to prefer doing a qualified import, which I would argue is confusing and should probably be changed (but it doesn't need to happen in this PR since it requires more thoughts)

src/Pirouette/Transformations/Defunctionalization.hs
```diff
@@ -351,13 +351,13 @@ rewriteHofType = go 0
...
-  where
-    go ty@SystF.TyFun {} = closureType ty
-    go x = x
+-- -- | This is a little more eager than 'rewriteHofType', in this case, we rewrite ALL
+-- -- function types.
+-- rewriteFunType :: (Language lang) => B.Type lang -> B.Type lang
+-- rewriteFunType = transform go
+--   where
+--     go ty@SystF.TyFun {} = closureType ty
+--     go x = x
```
aspiwack, line 360:
```suggestion
```

src/Pirouette/SMT/Monadic.hs
qaristote: Ok ! I saw some dead code [elsewhere](https://github.com/tweag/pirouette/blob/main/src/Pirouette/Transformations/Utils.hs#L25-L70) so I figured it was accepted within Pirouette.

src/Pirouette/Term/Syntax/Base.hs
qaristote: Yes exactly, `Data.Kind.Type` conflicts with `Pirouette.Term.Syntax.Base.Type` when imported unqualified.

0xd34df00d: As a compromise, to avoid importing either of `Type`s qualified or renaming the pirouette's `Type`, can we just omit `:: *` here and below? I think ghc will infer the right kind here.

0xd34df00d (approved):

src/Pirouette/Term/Syntax/Base.hs
aspiwack: Indeed, it's the default value.

PR #154, opened 2022-10-11
speed up smt calls

qaristote:
# Why ?

Currently, Pirouette uses SMT solvers by [calling them as external processes](https://github.com/tweag/pirouette/blob/3805638e0e11b5a0fc5cc6a41bd415d52674d23a/src/PureSMT/Process.hs#L26-L45) and [communicating with them using OS-level channels](https://github.com/tweag/pirouette/blob/3805638e0e11b5a0fc5cc6a41bd415d52674d23a/src/PureSMT/Process.hs#L47-L67). The language used to communicate with the solvers is SMTLib, which means the interaction goes as follows:
- Pirouette converts an s-expression (its internal representation for the solver's state) to an SMTLib command;
- it sends this command on a channel;
- the solver reads the command and parses it;
- it evaluates the result and converts it again from its internal representation to SMTLib;
- it sends this SMTLib output on a unix channel which is read by Pirouette;
- Pirouette converts the SMTLib output to an s-expression again.
This is considered slow, especially in comparison to directly linking the solvers' APIs inside Pirouette's code.

Some benchmarks verifying this intuition are available and detailed [here](/tweag/pirouette-benchmarks). Overall, here are the key takeaways:
-  on the biggest example of the test suite, Z3 is 2 times faster than the CVC instance used by default in Pirouette;
- using Z3 bindings instead of calling an external process speeds up the call to Z3 inside Pirouette by 14%;
- using bytestrings on top of bindings for the call to Z3 increases the speed-up to 14%;
- overall these two changes combined speed up Pirouette itself by 14%;
- these speed-ups could have been expected to be more important, but it is likely that Pirouette doing work in between each call to the solver hinders it from going at full speed;
- on the biggest example in the test suite, creating and parsing SMTLib commands takes up 12% of Pirouette running-time pre-speedup and 18% of its running-time post-speedup.

# What ?

This PR thus replaces the internals of the `PureSMT` module to call Z3 bindings instead of running solvers as external processes. We only use the [`z3_eval_smtlib2_string`](https://z3prover.github.io/api/html/group__capi.html#gaa623f645ba1458d1d4be9e023c05ae17) binding (which has Z3 evaluate SMTLib commands) instead of using bindings to manually build the AST because the serializing and parsing of SMTLib commands is deemed quick enough that the work needed to use these finer-grained bindings would not be worth the speedup. We thus also have these SMTLib commands be given to the solver as bytestrings instead of higher-level strings.

Commit c79d8bb4a4beb9ad448c4cf97130ebbfd9a068be replaces the internals of `PureSMT` to use bindings instead of running external processes. This introduces a dependency to the [`haskell-z3`](https://hackage.haskell.org/package/z3) library which ports Z3's C bindings to Haskell. As the library does not export the `z3_eval_smtlib2_string`, we actually rely on a [temporary fork](/qaristote/haskell-z3) that does.

This results in some of the `PureSMT` interface becoming useless as it was designed with running external processes in mind. Commit 64f18eb12229e7025596a59cb83ed07e03337fe0 cleans this interface.

As relying on a temporary fork of `haskell-z3` is not ideal, we instead use the bindings inside inlined C code in commit 33977c2e05fd32213a8c804a0eaf1329d8b3d7ef. There is still a dependency on (upstream) `haskell-z3` because I did not manage to get linking to work without it.

# What's next ?

Once the `haskell-z3` fork is [merged upstream](/IagoAbal/haskell-z3/pull/84) it shouldn't be necessary to inline C code anymore. Ideally `haskell-z3` should also be improved to allow using bytestrings in its higher-level modules `Z3.Monad` and `Z3.Base`, as currently only the lower-level `Z3.Base.C` allows it.  

This PR makes Pirouette not solver-universal anymore: the use of Z3 is forced. Other solvers could be supported again given that they provide an API with a function similar to `z3_eval_smtlib2_string`, but as of writing this is not the case of CVC4 or CVC5 for example.

Interestingly, our benchmarks also showed that even Z3's C API is not as fast as Z3 itself, most likely because `z3_eval_smtlib2_string` does not wrap Z3's internals the same way Z3's frontend does. This could be an additional direction in which to look to further improve the running-time of the interaction with Z3 inside Pirouette.

After this PR, serializing and parsing SMTLib commands will take up 18% of Pirouette's running time on the biggest example of the test suite. If this were to be deemed enough to decide that the internals of `PureSMT` should build the AST themselves (fully using Z3's API) instead of relying on SMTLib, recovering solver-universality would require a Haskell library akin to [`smt-switch`](/stanford-centaur/smt-switch) or [`metaSMT`](https://github.com/agra-uni-bremen/metaSMT) (which would likely be a lot of work).

### DIFF
--- a/nix/ci.nix
+++ b/nix/ci.nix
@@ -4,5 +4,6 @@
 let
   ourpkgs = import ./packages.nix {};
 in pkgs.mkShell {
+    nativeBuildInputs = ourpkgs.native-deps;
     buildInputs = ourpkgs.build-deps;
 }
--- a/nix/ci.nix
+++ b/nix/ci.nix
@@ -6,4 +6,5 @@ let
 in pkgs.mkShell {
     nativeBuildInputs = ourpkgs.native-deps;
     buildInputs = ourpkgs.build-deps;
+    LD_LIBRARY_PATH = "${pkgs.z3.lib}/lib";
 }
--- a/nix/ci.nix
+++ b/nix/ci.nix
@@ -6,5 +6,4 @@ let
 in pkgs.mkShell {
     nativeBuildInputs = ourpkgs.native-deps;
     buildInputs = ourpkgs.build-deps;
-    LD_LIBRARY_PATH = "${pkgs.z3.lib}/lib";
 }
--- a/package.yaml
+++ b/package.yaml
@@ -33,8 +33,7 @@ dependencies:
   - optics-core
   - optics-th
   # PureSMT deps
-  - process
-  - typed-process
+  - inline-c
   - z3
 
 library:
--- a/package.yaml
+++ b/package.yaml
@@ -34,13 +34,24 @@ dependencies:
   - optics-th
   # PureSMT deps
   - inline-c
-  - z3
+
 
 library:
   source-dirs: src
   ghc-options:
     -Wall
     -Wno-orphans
+  # Copied from https://hackage.haskell.org/package/z3-408.2/src/z3.cabal
+  #
+  # In OS X libz3 is linked statically against libgomp.
+  # In Windows Z3 is compiled using MS C++.
+  when:
+    - condition: os(darwin) || os(windows)
+      then:
+        extra-libraries: z3
+      else:
+        extra-libraries: gomp z3 gomp
+
 
 tests:
   spec:
@@ -64,4 +75,3 @@ tests:
 ##       -fprof-auto
 ##       -fexternal-interpreter
 ##       "-with-rtsopts=-N -p -s -h"
-
--- a/package.yaml
+++ b/package.yaml
@@ -35,6 +35,7 @@ dependencies:
   # PureSMT deps
   - process
   - typed-process
+  - z3
 
 library:
   source-dirs: src
--- a/src/PureSMT/Process.hs
+++ b/src/PureSMT/Process.hs
@@ -6,6 +6,7 @@ module PureSMT.Process where
 
 import Control.Monad
 import qualified Data.ByteString.Char8 as BS
+import qualified Data.ByteString.Unsafe as BS
 import Foreign.Ptr (Ptr)
 import qualified Language.C.Inline as C
 import qualified Language.C.Inline.Unsafe as CU
@@ -53,7 +54,7 @@ command solver cmd = do
     [CU.exp| const char* {
                   Z3_eval_smtlib2_string($(Z3_context ctx), $bs-cstr:cmdTxt)
                   } |]
-      >>= BS.packCString
+      >>= BS.unsafePackCString
   case readSExpr resp of
     Nothing -> do
       fail $ "solver replied with:\n" ++ BS.unpack resp
--- a/src/PureSMT/Process.hs
+++ b/src/PureSMT/Process.hs
@@ -1,73 +1,64 @@
+{-# LANGUAGE OverloadedStrings #-}
+
 module PureSMT.Process where
 
 import Control.Monad
-import Data.Maybe (fromMaybe)
-import Data.String (fromString)
+import qualified Data.ByteString.Char8 as BS
+import Data.Function ((&))
+import Foreign.Ptr (Ptr)
 import PureSMT.SExpr
 import System.IO
-import System.Mem.Weak
-import qualified System.Process as P
-import System.Process.Typed
+-- We use the low-level C API so that we can directly feed bytestrings to Z3
+-- TODO: add high-level functions using bytestrings to haskell-z3
+import Z3.Base.C
+  ( Z3_context,
+    z3_eval_smtlib2_string,
+    z3_mk_config,
+    z3_mk_context,
+  )
 import Prelude hiding (const)
 
-type SolverProcess = Process Handle Handle Handle
-
 data Solver = Solver
-  { process :: SolverProcess,
+  { context :: Ptr Z3_context,
     debugMode :: Bool
   }
 
-solverPid :: Solver -> IO (Maybe P.Pid)
-solverPid = P.getPid . unsafeProcessHandle . process
+-- TODO: rename this function and change its type
 
-unsafeSolverPid :: Solver -> IO P.Pid
-unsafeSolverPid = fmap (fromMaybe $ error "ProcessHandle was already closed") . solverPid
-
+-- | Create a brand-new context for Z3 to work in.
 launchSolverWithFinalizer ::
-  -- | Command to call the solver, will be made into a 'ProcessConfig' with 'fromString'
+  -- | Unused argument existing to match the old interface
   String ->
   -- | Whether or not to debug the interaction
   Bool ->
   IO Solver
-launchSolverWithFinalizer cmd dbg = do
-  p <- startProcess config
+launchSolverWithFinalizer _ dbg = do
+  solverCfg <- z3_mk_config
+  solverCtx <- z3_mk_context solverCfg
 
-  let s = Solver p dbg
+  let s = Solver solverCtx dbg
   setOption s ":print-success" "true"
   setOption s ":produce-models" "true"
 
-  -- Registers a finalizer to send an "exit" command to the solver, in case it
-  -- hasn't terminated yet. If it has terminated, we do nothing.
-  addFinalizer p (getExitCode p >>= maybe (send s (List [Atom "exit"])) (\_ -> return ()))
   return s
-  where
-    config :: ProcessConfig Handle Handle Handle
-    config = setStdin createPipe $ setStdout createPipe $ setStderr createPipe $ fromString cmd
 
-send :: Solver -> SExpr -> IO ()
-send solver cmd = do
-  let cmdTxt = showsSExpr cmd ""
+-- | Have Z3 evaluate a command in SExpr format.
+command :: Solver -> SExpr -> IO SExpr
+command solver cmd = do
+  let cmdTxt = serializeSExpr cmd
   when (debugMode solver) $ do
-    pid <- unsafeSolverPid solver
-    putStrLn ("[send: " ++ show pid ++ "] " ++ cmdTxt)
-  hPutStrLn (getStdin $ process solver) cmdTxt
-  hFlush (getStdin $ process solver)
-
-recv :: Solver -> IO SExpr
-recv solver = do
-  resp <- hGetLine (getStdout $ process solver)
+    BS.putStrLn $ "[send] " `BS.append` cmdTxt
+  resp <-
+    z3_eval_smtlib2_string (context solver)
+      & BS.useAsCString cmdTxt
+      >>= BS.packCString
   case readSExpr resp of
     Nothing -> do
-      rest <- hGetContents (getStdout $ process solver)
-      fail $ "solver replied with:\n" ++ resp ++ "\n" ++ rest
+      fail $ "solver replied with:\n" ++ BS.unpack resp -- ++ "\n" ++ rest
     Just (sexpr, _) -> do
-      pid <- unsafeSolverPid solver
       when (debugMode solver && sexpr /= Atom "success") $ do
-        putStrLn ("[recv: " ++ show pid ++ "] " ++ showsSExpr sexpr "")
+        putStrLn $ "[recv] " ++ showsSExpr sexpr ""
       return sexpr
-
-command :: Solver -> SExpr -> IO SExpr
-command solver cmd = send solver cmd >> recv solver
 
 -- | A command with no interesting result.
 ackCommand :: Solver -> SExpr -> IO ()
--- a/src/PureSMT/Process.hs
+++ b/src/PureSMT/Process.hs
@@ -55,7 +55,7 @@ command solver cmd = do
       >>= BS.packCString
   case readSExpr resp of
     Nothing -> do
-      fail $ "solver replied with:\n" ++ BS.unpack resp -- ++ "\n" ++ rest
+      fail $ "solver replied with:\n" ++ BS.unpack resp
     Just (sexpr, _) -> do
       when (debugMode solver && sexpr /= Atom "success") $ do
         putStrLn $ "[recv] " ++ showsSExpr sexpr ""
--- a/src/PureSMT/Process.hs
+++ b/src/PureSMT/Process.hs
@@ -7,7 +7,8 @@ module PureSMT.Process where
 import Control.Monad
 import qualified Data.ByteString.Char8 as BS
 import Foreign.Ptr (Ptr)
-import qualified Language.C.Inline as C
+import qualified Language.C.Inline as C hiding (block, exp)
+import qualified Language.C.Inline.Unsafe as C
 import PureSMT.SExpr
 import qualified PureSMT.Z3 as Z3
 import System.IO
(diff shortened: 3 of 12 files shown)


src/PureSMT/SExpr.hs
```diff
@@ -186,26 +210,33 @@ ppSExpr = go 0
...
-readSExpr (';' : more) = readSExpr $ drop 1 $ dropWhile (/= '\n') more
-readSExpr ('|' : more) = do
-  (sym, '|' : rest) <- pure (span ('|' /=) more)
-  Just (Atom ('|' : sym ++ ['|']), rest)
-readSExpr ('(' : more) = do
-  (xs, more1) <- list more
-  return (List xs, more1)
+readSExpr :: BS.ByteString -> Maybe (SExpr, BS.ByteString)
+readSExpr (c :< more) | isSpace c = readSExpr more
+readSExpr (';' :< more) = BS.dropWhile (/= '\n') more & BS.drop 1 & readSExpr
```
0xd34df00d, line 221: Very minor thing, but I'd vote for rewriting it in the other direction with `$` — `$` is used in the other branches, and I'd say it's more common in Haskell code.

0xd34df00d (approved):

0xd34df00d:
Perhaps somebody else wants to look at this though before it gets merged.

src/PureSMT/Process.hs
```diff
@@ -1,74 +1,66 @@
...
+  resp <-
+    [C.exp| const char* {
+                  Z3_eval_smtlib2_string($(Z3_context ctx), $bs-cstr:cmdTxt)
+                  } |]
+      >>= BS.packCString
   case readSExpr resp of
     Nothing -> do
-      rest <- hGetContents (getStdout $ process solver)
-      fail $ "solver replied with:\n" ++ resp ++ "\n" ++ rest
+      fail $ "solver replied with:\n" ++ BS.unpack resp -- ++ "\n" ++ rest
```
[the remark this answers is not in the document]
qaristote: Yep. Note that I'm not adding a line break at the end because Z3 already does it.

src/PureSMT/SExpr.hs
qaristote: Yeah I like `&`, `<&>` and `-.` way better but I guess that's just a me thing :'(

src/PureSMT/Process.hs
```diff
@@ -1,74 +1,66 @@
...
-  hPutStrLn (getStdin $ process solver) cmdTxt
-  hFlush (getStdin $ process solver)
-
-recv :: Solver -> IO SExpr
-recv solver = do
-  resp <- hGetLine (getStdout $ process solver)
+    BS.putStrLn $ "[send] " `BS.append` cmdTxt
+  let ctx = context solver
+  resp <-
+    [C.exp| const char* {
```
facundominguez, line 52: I think you want to use [these](https://hackage.haskell.org/package/inline-c-0.9.1.6/docs/Language-C-Inline-Unsafe.html) instead of the safe calls.

facundominguez (changes requested):
I would like the dependency on haskell-z3 to be removed before merging this. But otherwise it is looking good.

I suggested in a comment using unsafe FFI calls which could still provide a further speed up.

qaristote:
> I would like the dependency on haskell-z3 to be removed before merging this. 

I'm all for this but I've tried everything I could think of to compile the code without having `haskell-z3` as a dependency and I always get this linking error:
```
GHC.Runtime.Linker.dynLoadObjs: Loading temp shared object failed
During interactive linking, GHCi couldn't find the following symbol:
  /tmp/ghc237807_0/libghc_3.so: undefined symbol: Z3_mk_config
This may be due to you not asking GHCi to load extra object files,
archives or DLLs needed by your current session.  Restart GHCi, specifying
the missing library using the -L/path/to/object/dir and -lmissinglibname
flags, or simply by naming the relevant files on the GHCi command line.
Alternatively, this link failure might indicate a bug in GHCi.
If you suspect the latter, please report this as a GHC bug:
  https://www.haskell.org/ghc/reportabug
```

facundominguez:
I'll have a look.

src/PureSMT/Process.hs
```diff
@@ -1,74 +1,67 @@
...
+
 module PureSMT.Process where
 
 import Control.Monad
-import Data.Maybe (fromMaybe)
-import Data.String (fromString)
+import qualified Data.ByteString.Char8 as BS
+import Foreign.Ptr (Ptr)
+import qualified Language.C.Inline as C hiding (block, exp)
+import qualified Language.C.Inline.Unsafe as C
```
facundominguez, line 11: I would import this as `CU`, as the documentation of `Language.C.Inline.Unsafe` suggests. Then it isn't necessary to hide `block` and `exp` in the import of `Language.C.Inline`.

src/PureSMT/Process.hs
```diff
@@ -1,74 +1,67 @@
...
-recv :: Solver -> IO SExpr
-recv solver = do
-  resp <- hGetLine (getStdout $ process solver)
+    BS.putStrLn $ "[send] " `BS.append` cmdTxt
+  let ctx = context solver
+  resp <-
+    [C.exp| const char* {
+                  Z3_eval_smtlib2_string($(Z3_context ctx), $bs-cstr:cmdTxt)
+                  } |]
+      >>= BS.packCString
```
facundominguez, line 56: This is leaking the output bytestring after packing it, right?

In https://hackage.haskell.org/package/bytestring-0.11.3.1/docs/Data-ByteString-Unsafe.html#g:3 there are functions unsafePackCString and unsafePackCStringFinalizer that could be helpful.

facundominguez:
Now the interpreter is able to find `libz3.so` if `pirouette` is build inside the nix shell, which sets `LD_LIBRARY_PATH` appropriately.

package.yaml
```diff
@@ -33,8 +33,8 @@ dependencies:
   - optics-core
   - optics-th
   # PureSMT deps
-  - process
-  - typed-process
+  - inline-c
+  - z3
```
qaristote, line 37: This line should be removed as well if we truly want to remove the dependency on `haskell-z3`, otherwise `test/run-tests.sh` will repopulate `pirouette.cabal` with the dependency. But when I do so the build fails again with
```
Configuring test suite 'spec' for pirouette-2.0.0..
Preprocessing test suite 'spec' for pirouette-2.0.0..
Building test suite 'spec' for pirouette-2.0.0..
[ 1 of 13] Compiling Language.Pirouette.ExampleSpec ( tests/unit/Language/Pirouette/ExampleSpec.hs, /home/qaristote/code/pirouette/dist-newstyle/build/x86_64-linux/ghc-9.0.2/pirouette-2.0.0/t/spec/build/spec/spec-tmp/Language/Pirouette/ExampleSpec.o, /home/qaristote/code/pirouette/dist-newstyle/build/x86_64-linux/ghc-9.0.2/pirouette-2.0.0/t/spec/build/spec/spec-tmp/Language/Pirouette/ExampleSpec.dyn_o )
<command line>: /home/qaristote/code/pirouette/dist-newstyle/build/x86_64-linux/ghc-9.0.2/pirouette-2.0.0/build/libHSpirouette-2.0.0-inplace-ghc9.0.2.so: undefined symbol: Z3_mk_config
[FAILURE] 'cabal build' failed, will exit with 1
```

nix/ci.nix
```diff
@@ -6,5 +6,4 @@ let
 in pkgs.mkShell {
     nativeBuildInputs = ourpkgs.native-deps;
     buildInputs = ourpkgs.build-deps;
-    LD_LIBRARY_PATH = "${pkgs.z3.lib}/lib";
```
qaristote: Explicitly setting `LD_LIBRARY_PATH` isn't necessary because this is already dealt with by setting `nativeBuildInputs`. Note I used `nativeBuildInputs` because that's what usually works for me but I'm not sure this is the exact right `*buildInputs` attribute that should be used as [the documentation](https://nixos.org/manual/nixpkgs/stable/#variables-specifying-dependencies) is quite technical.

src/PureSMT/Process.hs
qaristote: Yeah indeed I hadn't thought of that. So I tried doing 
```
[CU.exp| char* {
                  // the function outputs a const char*
                  (char*) Z3_eval_smtlib2_string($(Z3_context ctx), $bs-cstr:cmdTxt)
                  } |]
>>= BS.unsafePackMallocCString
```
because calling `free` on the CString is the only finalizer I can think of (Z3 doesn't have a `del_string` function in its API because `Z3_string`s are just `const char*`s) and you can't free `const`s. But the tests fail at run-time with
```
  Symbolic
    Symbolic prove
      incorrectness triples
        [input > 0] add 1 [result > 0] counter:                                                           OK (0.01s)
        [input > 0] add 1 [result > 1] verified:                                                          free(): invalid size
./tests/run-tests.sh: line 57: 1297697 Aborted                 (core dumped) cabal run tests
[FAILURE] 'cabal run tests' failed; check the respective artifact.
```
I assume the `free() : invalid size` means it's trying to free a non-allocated pointer, but I don't see how it could happen when the first example works perfectly fine.

qaristote: Also if we're worried about memory leaks then the context also gets leaked doesn't it ? We don't send `(exit)` anymore so Z3 can't possibly know it should call `Z3_del_context`.

facundominguez: > Also if we're worried about memory leaks then the context also gets leaked doesn't it ?

That's a good point. In the case of the context, I'm guessing it is created only once per execution. But if pirouette were used as a library, perhaps it could be leaked multiple times.

The queries to the SMT solver do happen enough times that the leaks could be problematic in sufficiently large examples.

> but I don't see how it could happen when the first example works perfectly fine.

I went to check how the `Z3_string` is created. `Z3_eval_smtlib2_string` calls in turn to a function `mk_external_string` which returns a reference to a buffer kept in the context. If the reference is released, most likely the next call to Z3 that tries to use the buffer is going to fail.

In short, it looks like Z3 doesn't expect us to release the `Z3_string`s. I also checked [one of the examples](https://github.com/Z3Prover/z3/blob/fe1e3017354768e7761ad1db0805fb54b5d4f333/examples/c/test_capi.c) and it doesn't release `Z3_string`s either. So I'm a bit puzzled, how can we have more than one `Z3_string` at once if they both are pointers to the same buffer? There seem to be API calls which take more than one `Z3_string` as parameter.

facundominguez (approved):

src/PureSMT/Process.hs
qaristote: So in the end I should use `unsafePackCString` instead of `packCString` ? In both cases the underlying CString is not garbage collected but in the safe case the CString is copied to create the bytestring so it should be slower than the unsafe case which just computes the length of the string, right ?

facundominguez: Given that a later call to `mk_external_string` could invalidate previously returned Z3 strings, I think the safest option is to immediately copy the CString as your original code does.

src/PureSMT/Process.hs
```diff
@@ -43,6 +43,9 @@ initZ3instance dbg = do
 
   return s
 
+freeZ3Instance :: Solver -> IO ()
+freeZ3Instance s = void $ command s $ List [Atom "exit"]
```
facundominguez, line 47: Nobody seems to be calling this function.

Also, apparently destroying the context requires a call to `Z3_del_context`. I don't know if sending the exit command is necessary.

facundominguez:
I contributed one minor optimization. cc6d9b5 causes `inline-c` to use `unsafeUseAsCString` to marshal the cmdText to C. This can be observed with `-ddump-splices`.

src/PureSMT/Process.hs
qaristote: Given that the output bytestring is parsed and consumed (by `readSExpr`) inside the body of `command`, the unsafeness will only lead to an error if two instances of `command` are run concurrently and with the same solver instance (I'm assuming all of Pirouette's interaction with Z3 is done inside `command`). But Z3 is only considered thread-safe if the different threads use different contexts so I think it makes sense to keep this.

src/PureSMT/Process.hs
qaristote: Yeah I figured Z3 would call `Z3_del_context` by itself when evaluating `(exit)` but it was just a guess.

qaristote: I tried using
```
let ctx = context solver
 in [CU.exp| void { Z3_del_context($(Z3_context ctx)) } |]
```
but I'm getting a segmentation fault at runtime. I tried looking into what `(exit)` does but the [SMTLib documentation](https://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2021-05-12.pdf#subsection.4.2.1) doesn't say much. I think it's fair to stick with `(exit)` at least for now: in the worst case `freeZ3Instance` just won't do anything and the memory leak won't be a serious problem because Pirouette is not used as a library yet.

qaristote: Ok so 
```
printf("%s", Z3_eval_smtlib2_string(ctx, "(assert (< 0 -1))"));
printf("%s", Z3_eval_smtlib2_string(ctx, "(check-sat)"));
printf("%s", Z3_eval_smtlib2_string(ctx, "(exit)"));
printf("%s", Z3_eval_smtlib2_string(ctx, "(check-sat)"));
```
outputs
```
unsat
unsat
```
so I guess `(exit)` doesn't do anything here.

src/PureSMT/Process.hs
facundominguez: > Given that the output bytestring is parsed and consumed (by readSExpr) inside the body of command

I would agree with that, if `command` deepseq'ed the resulting SExp before returning it to the caller. As it stands, some suspended computation in the strings inside the SExp could leak a reference to the bytestring.

facundominguez: Also, I would hope that copying the bytestring is faster than deepseq'ing, but you might disprove me.

src/PureSMT/Process.hs
facundominguez: > but I'm getting a segmentation fault at runtime.

This looks to me the result of [src/PureSMT.hs](https://github.com/tweag/pirouette/blob/219689f1d5ecd9b487f40ffa85fe10a1398bc195/src/PureSMT.hs#L88) reusing the context after it has been destroyed.

Also, the use of `unsafePackCString` could interfere with this too.

src/PureSMT/Process.hs
qaristote: Oh yeah I didn't consider that. It would still be fine if the output SExpr were consumed before sending the next command but that wouldn't be compatible with exploring several branches in parallel.
And well I would guess copying bytestrings would make the overall amount of work bigger than just deepseq'ing, but that would only be true under the assumption that Pirouette always uses the output SExpr and I imagine it's not necessary when the command is something that will just return `success`. I'll just revert to `packCString` then.

src/PureSMT/Process.hs
qaristote: I mean `solveOpts` is pushing back the solver instance on the MStack after it has solved the problem so this could lead to this problem if the output function of type `Problem domain res -> res` were called several times, but I don't see why this would happen.

qaristote: But if that can be a problem I guess trying to free the solver instance is out of the question so I'll just give up on it for now.

Issue #156, opened 2022-10-21
Z3's logical context is leaked


qaristote:
Following (and as discussed in) #154, Pirouette does its SMT solving by [inline-C-calling a binding from Z3's API](https://github.com/tweag/pirouette/blob/748cbb68358061b234dc93452668e3de58772940/src/PureSMT/Process.hs#L57-L59), `Z3_eval_smtlib2_string`. This function expects a `Z3_context`, a C struct which we deal with in Haskell by [maintaining a pointer](https://github.com/tweag/pirouette/blob/748cbb68358061b234dc93452668e3de58772940/src/PureSMT/Process.hs#L21). As such, _it is not handled by Haskell's GC_: only the pointer is garbage-collected. It must thus be freed manually, be inline-C-calling the binding `Z3_del_context`.

As of now, this is simply not taken care of, and the `Z3_context` is leaked. This is not a big deal because Pirouette is currently used so that a solver instance is only created a few times during a run, hence only a few of these contexts get leaked before the program terminates and the memory is effectively freed. But this could become a problem if Pirouette were used as a library and a lot of solver instances were created during the same run.

As discussed [here](https://github.com/tweag/pirouette/pull/154#discussion_r997979440), freeing the context [inside the body of the `solveOpts` function](https://github.com/tweag/pirouette/blob/748cbb68358061b234dc93452668e3de58772940/src/PureSMT.hs#L85-L86) after the problem has been solved was tried, but it leads to a segmentation fault.

qaristote:
> As discussed [here](https://github.com/tweag/pirouette/pull/154#discussion_r997979440), freeing the context [inside the body of the `solveOpts` function](https://github.com/tweag/pirouette/blob/748cbb68358061b234dc93452668e3de58772940/src/PureSMT.hs#L85-L86) after the problem has been solved was tried, but it leads to a segmentation fault.

It looks like the reason this doesn't work is that the result of `solveOpts`, a function of type `Problem domain res -> res` which comes with a Z3 instance attached internally, [is used on two different problems](https://github.com/tweag/pirouette/blob/748cbb68358061b234dc93452668e3de58772940/src/Pirouette/Symbolic/Eval.hs#L184-L191): one generated by `CheckPath` and one generated by `CheckProperty`. Am I correct to assume this is the only way this output function is used more than once ?

[...]

PR #157, opened 2022-10-21
fix memory leak of Z3 context

qaristote:
As discussed in #156, following #154 the Z3 logical contexts are never garbage collected. This PR fixes this by adding a finalizer to the logical context pointer.

### DIFF
--- a/src/PureSMT/Process.hs
+++ b/src/PureSMT/Process.hs
@@ -11,7 +11,6 @@ import qualified Language.C.Inline as C
 import qualified Language.C.Inline.Unsafe as CU
 import PureSMT.SExpr
 import qualified PureSMT.Z3 as Z3
-import System.IO
 import Prelude hiding (const)
 
 C.context (C.baseCtx <> C.fptrCtx <> C.bsCtx <> Z3.cContext)
--- a/src/PureSMT/Process.hs
+++ b/src/PureSMT/Process.hs
@@ -6,7 +6,7 @@ module PureSMT.Process where
 
 import Control.Monad
 import qualified Data.ByteString.Char8 as BS
-import Foreign.Ptr (Ptr)
+import Foreign.ForeignPtr
 import qualified Language.C.Inline as C
 import qualified Language.C.Inline.Unsafe as CU
 import PureSMT.SExpr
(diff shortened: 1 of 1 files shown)


src/PureSMT/Process.hs
```diff
@@ -30,14 +29,18 @@ initZ3Instance ::
   Bool ->
   IO Solver
 initZ3Instance dbg = do
+  let ctxFinalizer =
+        [C.funPtr| void free_context(Z3_context ctx) {
+                                      Z3_del_context(ctx);
+                                      } |]
```
qaristote, line 35: The code here is a bit redundant. Is there a better way to do this ?

qaristote:
I'm also thinking this might be the right time to add some tests, in particular one that calls the finalizer on a context and checks that it has indeed been garbage collected.

src/PureSMT/Process.hs
facundominguez: I think I would have tried
```haskell
let ctxFinalizer ctx = [CU.block| void { Z3_del_context($(Z3_context ctx)); } |]
```

src/PureSMT/Process.hs
```diff
@@ -30,14 +29,18 @@ initZ3Instance ::
...
   solverCtx <-
-    [CU.block| Z3_context {
+    newForeignPtr ctxFinalizer
```
facundominguez, line 37: The finalizer can run in a different thread that the one that created the context. But perhaps that's fine for Z3. When you test it, you would have to make sure to build the test with the multi-threaded runtime.

src/PureSMT/Process.hs
qaristote: > I think I would have tried
> 
> ```haskell
> let ctxFinalizer ctx = [CU.block| void { Z3_del_context($(Z3_context ctx)); } |]
> ```

It doesn't type:

```
Couldn't match type: GHC.Ptr.Ptr Z3.LogicalContext -> IO ()
                     with: GHC.Ptr.FunPtr (GHC.Ptr.Ptr Z3.LogicalContext -> IO ())
      Expected: FinalizerPtr Z3.LogicalContext
        Actual: GHC.Ptr.Ptr Z3.LogicalContext -> IO ()
```

qaristote: Another option that does type is
```haskell
ctxFinalizer <- [CU.exp| void (*fun_ptr)(Z3_context) { &Z3_del_context } |]
```
Any favorites ?

facundominguez: I think I like the original more. You could have used this instead:
https://www.stackage.org/haddock/lts-19.30/base/Foreign-Concurrent.html#v:newForeignPtr

But I don't think it is better in any obvious way.

facundominguez (approved):

src/PureSMT/Process.hs
facundominguez: Here there is a discussion that argues to keep the code as is: https://hackage.haskell.org/package/inline-c-0.9.1.6/docs/Language-C-Inline.html#v:funPtr

mmontin (approved):

PR #158, opened 2022-10-25
Flakifying Pirouette using haskell.nix

gabrielhdt:
Use `haskell.nix` to provide a proper build environment.

All content in `nix/` and `shell.nix` may be deleted if we want to force the migration to flakes.

The flake may be tested using `direnv` (a `.envrc` is set up), or simply with
```sh
$ nix develop
$ cabal run
```

Profiling has not been tested yet.

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -1,0 +1,6 @@
+module Main (main) where
+
+import Distribution.Simple
+
+main :: IO ()
+main = defaultMain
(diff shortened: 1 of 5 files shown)


facundominguez:
Is there some learning material about flakes for those of us not yet enlightened?

Also, do the build instructions need to be updated in the README?

gabrielhdt:
@facundominguez For flakes in general, I think the nixos manual is quite good <https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake.html>.

Indeed the build instructions ought to be updated.

PR #159, opened 2022-10-31
PureSMT: send commands with no output by batches

qaristote:
This PR merges together SMTLib2 commands whose output can be ignored (typically when the output is either `(success)` or an error). In practice, these commands are effectively added to a queue given by a [ByteString Builder](https://hackage.haskell.org/package/bytestring-0.11.3.1/docs/Data-ByteString-Builder.html) and this queue is only sent to the solver as one big batched command when the output of the next command cannot be ignored. This also means that in non-debug mode the commands are not checked to be valid and may thus lead to an error that isn't caught immediately.

The expected speed-up is described in the [`qa/z3_binding_faster_eval`](https://github.com/tweag/pirouette-benchmarks/commit/e09e0a7adc3b5260b3f2dbbae5f95a9dd679fad3) branch of [`pirouette-benchmarks`](https://github.com/tweag/pirouette-benchmarks): Pirouette is now 1.16 times faster on the largest example, `isUnity`. Note that these measurements were made with [a version of Z3](https://github.com/Z3Prover/z3/pull/6422) that improves the speed of the binding used in Pirouette's code but isn't part of a release yet, so until the next release the actual speed-up may differ.

While that wasn't measured to improve the speed of Pirouette, the serialization of s-expressions to bytestring is now done using the [untrimmed strategy](https://hackage.haskell.org/package/bytestring-0.11.3.1/docs/Data-ByteString-Builder-Extra.html#v:untrimmedStrategy) which is theoretically faster for building a bytestring which will be consumed immediately afterwards. Similarly, we manually set the size of the chunks depending on whether a single command or a whole batch is being serialized.

### DIFF
--- a/src/PureSMT/Process.hs
+++ b/src/PureSMT/Process.hs
@@ -43,43 +43,27 @@ initZ3Instance dbg = do
                      return ctx;
                      } |]
   solverQueue <- newIORef mempty
-  let s = Solver solverCtx dbg solverQueue
-  setOption s ":print-success" "true"
-  setOption s ":produce-models" "true"
-
-  return s
+  let solver = Solver solverCtx dbg solverQueue
+  when (debugMode solver) $ do
+    -- this should not be enabled in non-debug mode, as it messes with parsing
+    -- the outputs of commands that are actually interesting
+    setOption solver ":print-success" "true"
+  setOption solver ":produce-models" "true"
+  return solver
 
 -- | Send a bytestring to Z3.
 -- This function is thread-safe as long as concurrent instances do not share the
 -- same logical context.
-send :: Solver -> BS.ByteString -> IO (BS.ByteString)
+send :: Solver -> BS.ByteString -> IO SExpr
 send solver cmd = do
   let ctx = context solver
   when (debugMode solver) $ do
     BS.putStrLn $ "[send] " `BS.append` cmd
-  [CU.exp| const char* {
+  resp <-
+    [CU.exp| const char* {
          Z3_eval_smtlib2_string($fptr-ptr:(Z3_context ctx), $bs-ptr:cmd)
          } |]
-    >>= BS.packCString
-
--- | Force the solver to evaluate the queued commands.
--- The result is not checked to ensure the commands were evaluated correctly.
-flush :: Solver -> IO ()
-flush solver = do
-  cmds <- do
-    let solverQueue = queue solver
-    cmds <- readIORef solverQueue
-    writeIORef solverQueue mempty
-    return $ serializeBatch cmds
-  _ <- send solver cmds
-  return ()
-
--- | Have Z3 evaluate a command in SExpr format.
-command :: Solver -> SExpr -> IO SExpr
-command solver expr = do
-  flush solver
-  let cmd = serializeSingle $ renderSExpr expr
-  resp <- send solver cmd
+      >>= BS.packCString
   case readSExpr resp of
     Nothing -> do
       fail $ "solver replied with:\n" ++ BS.unpack resp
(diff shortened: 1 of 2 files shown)


src/PureSMT/SExpr.hs
```diff
@@ -62,14 +62,26 @@ overAtomS :: (String -> String) -> SExpr -> SExpr
...
+-- (the size of the buffer is expected to be small).
+serializeSingle :: Builder -> BS.ByteString
+serializeSingle = serializeUntrimmed 256 2048
+
+-- | Evaluate a bytestring builder corresponding to a batch of SMTLib2 commands
+-- (the size of the buffer is expected to be important).
+serializeBatch :: Builder -> BS.ByteString
+serializeBatch = serializeUntrimmed smallChunkSize defaultChunkSize
+
+-- | Convert an s-expression to a (strict) null-terminated bytestring.
```
facundominguez, line 80: This doesn't look null-terminated, and isn't a bytestring.

facundominguez (commented):
Looking good. Some questions and comments.

src/PureSMT/Process.hs
```diff
@@ -39,26 +42,44 @@ initZ3Instance dbg = do
...
+    return $ serializeBatch cmds
+  _ <- send solver cmds
+  return ()
+
+-- | Have Z3 evaluate a command in SExpr format.
+command :: Solver -> SExpr -> IO SExpr
+command solver expr = do
+  flush solver
+  let cmd = serializeSingle $ renderSExpr expr
+  resp <- send solver cmd
```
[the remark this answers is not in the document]
qaristote: Because I expect the queue to produce an output that is `(success)` x n which makes it harder to parse the result of the last command, which we're actually interested in.

src/PureSMT/SExpr.hs
```diff
@@ -62,14 +62,26 @@ overAtomS :: (String -> String) -> SExpr -> SExpr
...
-      "(" <> mconcat (intersperse " " [renderSExpr e | e <- es]) <> ")"
+-- | Evaluate a bytestring builder to a strict bytestring that is expected to be
+-- consumed immediately.
+serializeUntrimmed :: Int -> Int -> Builder -> BS.ByteString
+serializeUntrimmed firstChunkSize newChunksSize = LBS.toStrict . toLazyByteStringWith (untrimmedStrategy firstChunkSize newChunksSize) "\NUL"
+
+-- | Evaluate a bytestring builder corresponding to a single SMTLib2 command
+-- (the size of the buffer is expected to be small).
+serializeSingle :: Builder -> BS.ByteString
+serializeSingle = serializeUntrimmed 256 2048
```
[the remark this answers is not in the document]
qaristote: 256 is the first power of 2 that is bigger than the length of the longest command with interesting output in 'isUnity', and 2048 is just four times that because `smallChunkSize * 4 = defaultChunkSize`.

src/PureSMT/Process.hs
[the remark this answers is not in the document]
facundominguez: Ah, makes sense. Maybe you could mentioned that in a comment.

src/PureSMT/SExpr.hs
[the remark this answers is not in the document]
facundominguez: Thanks. Would be good to put it in a comment too.

src/PureSMT/Process.hs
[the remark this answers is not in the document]
qaristote: I think I'll just disable the `:print-success` option when not in debug mode.

src/PureSMT/Process.hs
```diff
@@ -88,13 +72,39 @@ command solver expr = do
         putStrLn $ "[recv] " ++ showsSExpr sexpr ""
       return sexpr
 
+-- | Push a command on the solver's queue of commands to evaluate.
+-- The command must not produce any output when evaluated.
+putQueue :: Solver -> SExpr -> IO ()
+putQueue solver expr = do
+  let solverQueue = queue solver
+  cmds <- readIORef solverQueue
+  writeIORef solverQueue $ cmds <> renderSExpr expr
```
facundominguez, line 81:
```haskell
putQueue solver expr =
  modifyIORef (queue solver) (<> renderSExpr expr)
```

src/PureSMT/Process.hs
```diff
@@ -18,7 +20,8 @@ C.include "z3.h"
 
 data Solver = Solver
   { context :: ForeignPtr Z3.LogicalContext,
-    debugMode :: Bool
+    debugMode :: Bool,
+    queue :: IORef Builder
```
facundominguez, line 24:
```suggestion
    queue :: IORef Builder   -- only used in non-debug mode
```

facundominguez (approved):
LGTM
Made a couple of suggestions.

src/PureSMT/Process.hs
qaristote: Neat ! But maybe `modifyIORef'` should be used here instead ?

> Be warned that [modifyIORef](https://hackage.haskell.org/package/base-4.17.0.0/docs/Data-IORef.html#v:modifyIORef) does not apply the function strictly. This means if the program calls [modifyIORef](https://hackage.haskell.org/package/base-4.17.0.0/docs/Data-IORef.html#v:modifyIORef) many times, but seldom uses the value, thunks will pile up in memory resulting in a space leak. This is a common mistake made when using an IORef as a counter. For example, the following will likely produce a stack overflow:
> `ref <- newIORef 0`
> `replicateM_ 1000000 $ modifyIORef ref (+1)`
> `readIORef ref >>= print` 
> To avoid this problem, use [modifyIORef'](https://hackage.haskell.org/package/base-4.17.0.0/docs/Data-IORef.html#v:modifyIORef-39-) instead.
>

qaristote: Yeah I guess not (`Builder f1 <> Builder f2 = Builder (f1 . f2)` is in WHNF). But then should we be concerned about the possibility that the queue becomes too big ?

facundominguez: Because this is a queue, and not something that produces a type with bounded memory (e.g. an `Int`), I think we could let it as is for now.

qaristote: It's a bit late but for a future PR it might be better to use `atomicModifyIORef`. It should be thread-safe (although not equivalent) to write two commands to a single queue concurrently. LiquidFixpoint actually does [something similar](https://github.com/ucsd-progsys/liquid-fixpoint/blob/544f8b0ba6d03b060701961250cce012412039c4/src/Language/Fixpoint/Smt/Interface.hs#L451-L460) (it puts a builder inside a TVar) for the sole sake of concurrently writing commands whose output can be discarded.

facundominguez: Using `atomicModifyIORef` could be necessary if multiple threads were using the same queue, but that its not the case, and I don't imagine it being the case soon either.

PR #162, opened 2022-12-16
Replace `PureSMT.Z3` with dependency on `smtlib-backends-z3`

qaristote:
`smtlib-backends` [is out](https://hackage.haskell.org/package/smtlib-backends). This PR thus replaces the `PureSMT.Z3` module and its inlined C code with a dependency on `smtlib-backends-z3`, whose implementation is similar.


src/PureSMT/Process.hs
```diff
@@ -5,119 +5,62 @@
...
-
--- | Have Z3 evaluate a command in SExpr format.
+launchSolver dbg = do
+  z3 <- Z3.toBackend <$> Z3.new
+  solver <- Bck.initSolver z3 lazy
+  return $ Solver solver dbg
+  where
+    lazy = dbg
+
+-- | Have the evaluate a command in SExpr format.
```
Niols, line 34:
```suggestion
-- | Have the solver evaluate a command in SExpr format.
```

src/PureSMT/Process.hs
```diff
@@ -5,119 +5,62 @@
...
 -- This forces the queued commands to be evaluated as well, but their results are
 -- *not* checked for correctness.
```
Niols, line 36: If I'm not mistaken, there are no queued commands anymore? Or at least not visible from PureSMT's point of view?

src/PureSMT/Process.hs
```diff
@@ -5,119 +5,62 @@
...
+  when (debugMode solver) $ do
+    hPutBuilder stdout $ "[send] " <> cmd <> "\n"
+    hFlush stdout
+  result <- Bck.command (backend solver) cmd
+  case readSExpr result of
+    Nothing -> do
+      fail $ "solver replied with:\n" ++ LBS.unpack result
+    Just (expr', _) -> do
+      when (debugMode solver) $ do
+        putStrLn $ "[recv] " ++ showsSExpr expr' ""
```
Niols, line 49: The other times we go for `hPutBuilder` + `hFlush` but here we use `putStrLn`. Is there a reason to that?

Niols (approved):

src/PureSMT/Process.hs
qaristote: They are still visible in that we manually enable queuing inside the `launchSolver` function, but we could decide to get rid of this comment yes.

src/PureSMT/Process.hs
qaristote: Not really, I guess it's just an artifact of which parts of the code I changed.

src/PureSMT/Process.hs
Niols: Is that the `lazy` parameter of `Bck.initSolver`? Otherwise, when reading the code of `launchSolver`, I don't really see that there is a queue somewhere.

qaristote: It was, I made its name more explicit in https://github.com/tweag/pirouette/pull/162/commits/80a850a02cc1953582ac616bbc16e0e21caa7d4a.

Niols: As well as fixing a bug, I see? Great!

Niols (approved):

Niols:
As far as I am concerned, this can be merged.

PR #163, opened 2022-12-23
Update build system

gabrielhdt:
- Use flakes to fix dependencies versions,
- Remove some unused dependencies that were used to build Plutus, but there is no need to build Plutus
- Update the cache mechanism of the CI

The continuous integration takes 1m34s to complete using the cache.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,3 +1,4 @@
+name: Build and test
 
 on:
   push:
@@ -7,8 +8,8 @@ on:
   pull_request:
 
 jobs:
-  build-and-test:
-    name: build-and-test
+  build-and-run-tests:
+    name: Build & run tests
     runs-on: ubuntu-latest
     steps:
     - name: Check out repository code.
@@ -44,16 +45,16 @@ jobs:
           ./*.res
 
   check-result:
-    needs: build-and-test
-    name: check-result
+    needs: build-and-run-tests
+    name: Check tests output
     runs-on: ubuntu-latest
     steps:
-    - name: Access the results
+    - name: Access tests output
       uses: actions/download-artifact@v3
       with:
         name: pirouette-checks
 
-    - name: Check result
+    - name: Check output
       run: |
         echo "!! CABAL TEST OUTPUT:"
         cat pirouette-cabal-test.out


Niols (changes requested):
LGTM, except:

- There is a leftover `default.nix` that cannot even work (because the directory `nix` doesn't exist anymore).

- Wouldn't it be worth using `haskell.nix` here? With such a simple `cabal.project`, it should be very efficient.

.github/workflows/build-and-test.yml
```diff
@@ -15,37 +15,33 @@ jobs:
       uses: actions/checkout@v2.4.0
 
     - name: Install nix
-      uses: cachix/install-nix-action@v15
+      uses: cachix/install-nix-action@v18
```
Niols, line 18: We could consider adding dependabot there.

.github/workflows/build-and-test.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         key: ${{ runner.os }}-${{ hashFiles('pirouette.cabal', 'cabal.project') }}]
         restore-keys: ${{ runner.os }}-
 
-    - name: Accessing the result cache
+    - name: Accessing the result artifact
```
Niols, line 35: My bad! I meant that currently we're using the cache to send files from one job to another (and therefore the name of the step was fine as it was), when it looks to me like we should go through the artifacts mechanism instead. It would also allow us to download the results of the tests if we need to debug.

Niols (approved):

.github/workflows/build-and-test.yml
Niols: Maybe we should just keep it as it is for now. It works, after all.

.github/workflows/build-and-test.yml
Niols: Done in [efeb3f9](https://github.com/tweag/pirouette/pull/163/commits/efeb3f9e6e0dccacd49501f8de3260485ac637b3).

.github/workflows/build-and-test.yml
Niols: There can be good reasons to split the CI in several jobs. My guess here is that it would allow triggering an upload of the Cabal cache even when tests fail, which can indeed be very valuable.

Niols: Documented in https://github.com/tweag/pirouette/issues/164.

.github/workflows/build-and-test.yml
```diff
@@ -15,37 +15,33 @@ jobs:
...
+        restore-keys: ${{ runner.os }}-
 
     - name: Accessing the result cache
-      uses: actions/cache@v2
+      uses: actions/cache@v3
       with:
         path: |
           ./*.out
           ./*.res
         key: result-cache-${{ github.sha }}
```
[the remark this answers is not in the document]
Niols: Followed in [6d0a06f](https://github.com/tweag/pirouette/pull/163/commits/6d0a06f2cd19f649f7513e9c6c3b0fe73440441e) and then https://github.com/tweag/pirouette/pull/163#discussion_r1060583773.

Niols:
While we're at it, would you mind giving a name to the workflow? Currently, GitHub's interface shows:
image
I suppose we could just call it “build-and-test”.

Niols:
All good for me, feel free to press squash that button!

Niols:
fixes #164

(I hope those closing keywords work everywhere but I doubt it.)

PR #168, opened 2023-01-17
Add a standard library for the example language

Niols:
This PR adds a tiny standard library for the example language. This allows shortening slightly the current tests. More importantly, it will help maximise sharing of usual functions in other tests, and I am planning to write quite a few of those in the future.

Technically, the example language does not support any notion of imports. Since the code of the tests would use those standard functions without defining them itself, the type checker would reject those pieces of code because they use undefined functions. One solution is to use the quasi-quoter `progNoTC` (read “program with no type-checking”), then to merge the resulting unchecked definitions with the ones from the standard library, and then only to call the type-checker. I attempted to make this slightly easier by providing a quasi-quoter `progWithStdLib` which basically does this under the hood. I don't know if that's the best design so I'm very open to remarks/suggestions.

Things that aren't that great yet:

- There is no notion of namespacing or modules, so all the names of the standard library are in the scope and that's it. This might become annoying if we want similar function names acting on different structures. Maybe we should consider prefixing all the functions in the standard library by the structure they act on. eg. `Maybe_map` vs. `List_map`, etc. (I'm not particularly attached to the underscore.)

- ~I don't currently type-check the standard library, meaning that the errors in it will only be found by and attributed to the quasi-quoter `progWithStdLib` when used somewhere else. This shouldn't be too hard to fix: one probably just want to run the type checker when creating the `stdLib` value.~ Edit: fixed in 1a06ec3300f7e5ca8cec025b4988e5eb7f2eeb90. Somehow, the output from Template Haskell is prettier and doesn't include the unreadable `Code:` part. But this is good enough IMHO.

- ~I don't provide a function other than `unionPrtUODefs` while only allows to combine definitions two by two, meaning we will have to call it a number of times linear in the number of “modules” (maybe a better word is “parts”) in the standard library. This is also pretty easy to fix.~ Edit: fixed in f6ffe07.

What do you guys think?

### DIFF
--- a/src/Language/Pirouette/Example/IsUnity.hs
+++ b/src/Language/Pirouette/Example/IsUnity.hs
@@ -59,19 +59,11 @@ pairEq @a @b eqA eqB x y =
     (\(x0 : a) (x1 : b) . match_Pair @a @b y @Bool
       (\(y0 : a) (y1 : b) . and (eqA x0 y0) (eqB x1 y1)))
 
-data Maybe x
-  = Just : x -> Maybe x
-  | Nothing : Maybe x
-
-maybeSum : forall x . Maybe x -> Maybe x -> Maybe x
-maybeSum @x mx my =
-  match_Maybe @x mx @(Maybe x)
-    (\(jx : x) . Just @x jx)
-    my
-
-isJust : forall x . Maybe x -> Bool
-isJust @x mx =
-  match_Maybe @x mx @Bool (\(jx : x) . True) False
+maybeSum : forall a . Maybe a -> Maybe a -> Maybe a
+maybeSum @a m1 m2 =
+  match_Maybe @a m1 @(Maybe a)
+    m2
+    (\(j1 : a) . Just @a j1)
 
 data KVMap k v
   = KV : List (Pair k v) -> KVMap k v
@@ -110,17 +102,15 @@ data TxInfo
 assetClassValueOf : Value -> Pair String String -> Integer
 assetClassValueOf v ac =
   match_Pair @String @String ac @Integer
-    (\(curSym : String) (tokName : String)
-     . match_Value v @Integer
-     (\(openV : KVMap String (KVMap String Integer))
-      . match_Maybe @(KVMap String Integer) (lkup @String @(KVMap String Integer) eqString openV curSym) @Integer
-          (\(tokM : KVMap String Integer)
-           . match_Maybe @Integer (lkup @String @Integer eqString tokM tokName) @Integer
-               (\(r : Integer) . r)
-               0
-          )
-          0
-    ))
+    (\(curSym : String) (tokName : String) .
+      match_Value v @Integer
+        (\(openV : KVMap String (KVMap String Integer)) .
+          match_Maybe @(KVMap String Integer) (lkup @String @(KVMap String Integer) eqString openV curSym) @Integer
+            0
+            (\(tokM : KVMap String Integer) .
+              match_Maybe @Integer (lkup @String @Integer eqString tokM tokName) @Integer
+                0
+                (\(r : Integer) . r))))
 
 -- Now we define the wrong isUnity function, that is too permissive
 wrong_isUnity : Value -> Pair String String -> Bool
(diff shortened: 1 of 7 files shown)


src/Language/Pirouette/Example/IsUnity.hs
```diff
@@ -93,19 +59,11 @@ pairEq @a @b eqA eqB x y =
...
-maybeSum : forall x . Maybe x -> Maybe x -> Maybe x
-maybeSum @x mx my =
-  match_Maybe @x mx @(Maybe x)
-    (\(jx : x) . Just @x jx)
-    my
-
-isJust : forall x . Maybe x -> Bool
-isJust @x mx =
-  match_Maybe @x mx @Bool (\(jx : x) . True) False
+maybeSum : forall a . Maybe a -> Maybe a -> Maybe a
```
0xd34df00d, line 62: This looks useful! Should this go to the stdlib as well?

src/Language/Pirouette/Example/IsUnity.hs
```diff
@@ -144,17 +102,15 @@ data TxInfo
...
-           . match_Maybe @Integer (lkup @String @Integer eqString tokM tokName) @Integer
-               (\(r : Integer) . r)
-               0
-          )
-          0
-    ))
+    (\(curSym : String) (tokName : String) .
+      match_Value v @Integer
+        (\(openV : KVMap String (KVMap String Integer)) .
+          match_Maybe @(KVMap String Integer) (lkup @String @(KVMap String Integer) eqString openV curSym) @Integer
```
0xd34df00d, line 108: Looks like `match_Maybe` args order has changed. Just curious, is it just handier this way or is there a deeper reason?

src/Language/Pirouette/Example/StdLib.hs
```diff
@@ -0,0 +1,116 @@
...
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE UndecidableInstances #-}
+-- -l has the effect of causing the TH interpreter to try loading
+-- z3 to resolve otherwise undefined symbols
+{-# OPTIONS_GHC -lz3 #-}
```
0xd34df00d, line 15: That's interesting! Does this module really depend on z3?

src/Language/Pirouette/QuasiQuoter/Internal.hs
```diff
@@ -0,0 +1,87 @@
...
+    { quoteExp = quote,
+      quotePat = \_ -> failure0 "patterns",
+      quoteType = \_ -> failure0 "types",
+      quoteDec = \_ -> failure0 "declarations"
+    }
+  where
+    failure0 k =
+      fail ("this quasiquoter does not support splicing " ++ k)
+
+-- Orphan instances to 'Lift' the necessary types
```
0xd34df00d, line 57: Would `{-# OPTIONS_GHC -Wno-orphans #-}` result in cleaner builds (eventually, even if this warning is disabled project-wide now)?

0xd34df00d (approved):

src/Language/Pirouette/Example/StdLib.hs
Niols: I must just have copied this from another file; my bad. 6e6f5f8e937e3254663eadf5a640a779c4c2a7e4 removes it a look just fine.

src/Language/Pirouette/Example/IsUnity.hs
Niols: I was doing things more on a define-by-need basis. But it's true that it looks rather generic. https://github.com/tweag/pirouette/pull/168/commits/0e079cbdd2715e920ad6378fd9480535e5812543 moves it to the standard library under the name `firstJust` (the only convincing name Hoogle threw at me when I asked for the type; I'm trying to stay as close as possible to Haskell type and function names).

src/Language/Pirouette/Example/IsUnity.hs
Niols: Not a very deep reason, I just try to follow how things are happening in Haskell and it looks like the prelude defines `Maybe a` as `Nothing | Just a` with `Nothing` first, so I went for that too. I also like it better personally, now that I see the difference, because usually the case of `Nothing` is fairly simple so you can read it in one go and get it out of the way and then only go read the harder case.

src/Language/Pirouette/QuasiQuoter/Internal.hs
Niols: Frankly, I am not a good enough Haskeller to have any idea whether it would improve the situation or not. I added it in https://github.com/tweag/pirouette/pull/168/commits/34d8e7930b43009ec299916ccc20f6ee57fec134.

Niols: This does raise another question for me: why did we disable the warning on a project level if it's easy to disable it only on some specific well-defined files?

src/Language/Pirouette/Example/IsUnity.hs
Niols: Does this answer convince you or would you rather go back to how it was before? I also don't have a very strong opinion and the type-checker will have our back anyways.

src/Language/Pirouette/QuasiQuoter/Internal.hs
0xd34df00d: > Frankly, I am not a good enough Haskeller to have any idea whether it would improve the situation or not.

Ideally, we'd define these instances in the module defining the types in question, especially if we expect them to be used in TH contexts.

> This does raise another question for me: why did we disable the warning on a project level if it's easy to disable it only on some specific well-defined files?

Dunno, that's a good question! I personally tend to disable only those warnings that I expect to occur in more or less every module in the project.

src/Language/Pirouette/Example/IsUnity.hs
Niols: I don't think you did!

PR #171, opened 2023-01-18
Add a second test about lists

Niols:
This PR builds on top of #170 and adds a second test. This one is here to defeat the symbolic engine with a big path explosion and an infinite recursion, which it does for now. I still need to figure out how to add fuel to this.


Niols:
And now with a simpler test that hands, while technically it should be possible to find something.

Niols:
It now strikes me that those tests might be more relevant as execution tests, rather than prover tests.

PR #174, opened 2023-01-19
Add an `hlint` pre-commit hook and cleanup files accordingly

### LINKED ISSUE #173: Add a `hlint` pre-commit hook and cleanup files accordingly.
cf https://github.com/tweag/pirouette/pull/172#issuecomment-1396840396

Niols:
closes #173 

builds on top of #172 

Now that I look at this, I am not convinced that we really want or need `hlint`. Maybe only for the unused pragmas?

### DIFF
--- a/src/Pirouette/Term/Syntax.hs
+++ b/src/Pirouette/Term/Syntax.hs
@@ -41,7 +41,7 @@ separateBoundFrom u t =
 
     -- `structuredInter` transforms the set of name clashes into a map
     -- from nameString to the list of nameUnique associated.
-    structuredInter inter = Set.fold (\n m -> Map.insertWith (++) (nameString n) [nameUnique n] m) Map.empty inter
+    structuredInter = Set.fold (\n m -> Map.insertWith (++) (nameString n) [nameUnique n] m) Map.empty
 
     -- `nextFresh s txt i n` takes a set `s` of names and a `nameString` `txt` and outputs
     -- the first index which does not create clash.
--- a/src/UnionFind/Monadic.hs
+++ b/src/UnionFind/Monadic.hs
@@ -105,8 +105,9 @@ find ::
   (Ord key, Monad m) =>
   key ->
   WithUnionFindT key value m (key, Int, Maybe value)
-find key =
-  Map.lookup key <$> getBindings >>= \case
+find key = do
+  bindings <- getBindings
+  case Map.lookup key bindings of
     Nothing -> return (key, 1, Nothing)
     Just (ChildOf key') -> do
       -- We use @findExn@ because invariant (ii) guarantees that it will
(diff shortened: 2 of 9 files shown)


0xd34df00d (approved):

src/Pirouette/Term/Syntax.hs
```diff
@@ -41,7 +41,7 @@ separateBoundFrom u t =
 
     -- `structuredInter` transforms the set of name clashes into a map
     -- from nameString to the list of nameUnique associated.
-    structuredInter inter = Set.fold (\n m -> Map.insertWith (++) (nameString n) [nameUnique n] m) Map.empty inter
+    structuredInter = Set.fold (\n m -> Map.insertWith (++) (nameString n) [nameUnique n] m) Map.empty
```
Niols, line 44: I am not necessarily a big fan of eta-reduction. I don't think it brings more clarity.

src/UnionFind/Monadic.hs
```diff
@@ -105,8 +105,9 @@ find ::
   (Ord key, Monad m) =>
   key ->
   WithUnionFindT key value m (key, Int, Maybe value)
-find key =
-  Map.lookup key <$> getBindings >>= \case
+find key = do
+  bindings <- getBindings
+  case Map.lookup key bindings of
     Nothing -> return (key, 1, Nothing)
```
Niols, line 111: I don't really think this change improves readability. It avoids the two monadic operators on the same line but I think it was already reasonably understandable before, and actually a bit more concise (not that concision should be the goal).

Niols (commented):
So, I'm not necessarily a big fan of the suggestions implemented in this PR. See in particular my two review comments. I also know of other examples (not in this project in particular) where `hlint`'s suggestions actually break the project, for instance by removing language pragmas that are actually used. I see three possible ways to go:

1. We can decide that it wasn't a great idea after all and not add `hlint`.
2. We can decide that we add all of `hlint` and implement all the changes in this PR.
3. We can decide that we like `hlint` but that we will ignore some warnings that we don't like this much.

WDYT @0xd34df00d @gabrielhdt?

src/UnionFind/Monadic.hs
Niols: FYI, the change implemented here is not what `hlint` suggested, but it does make `hlint` happy. `hlint`'s suggestion was:
image
which I find extremely ugly.

Niols: FYI, the change implemented here is not what `hlint` suggested, but it does make `hlint` happy. `hlint`'s suggestion was:
image
which I find extremely ugly.

Niols:
That sounds good to me but maybe we should ask the whole team what they think before taking such a decision.

Niols:
Nah, let's not.

PR #177, opened 2023-01-26
Add an `hpack` pre-commit hook

Niols:
Somehow, [the default `hpack` hook](https://github.com/cachix/pre-commit-hooks.nix/blob/7bdf85f6bbef581eb687838d19f2b35a4c9d77f0/modules/hooks.nix#L419-L435) does not work for us, so we write our own, which we call `hpack-sync` so as not to clash with the default. Only the first commit matters. The others will be there for testing purposes.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -124,11 +124,11 @@
         "nixpkgs-stable": "nixpkgs-stable"
       },
       "locked": {
-        "lastModified": 1674122161,
-        "narHash": "sha256-9QM4rvgUSEwO8DWtJN9sR/afEqrH1s3b6ACsZT5wiAM=",
+        "lastModified": 1675688762,
+        "narHash": "sha256-oit/SxMk0B380ASuztBGQLe8TttO1GJiXF8aZY9AYEc=",
         "owner": "cachix",
         "repo": "pre-commit-hooks.nix",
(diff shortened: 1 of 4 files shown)


Niols:
I marked it as draft because there are still things I do not understand about this `hpack` hook. Besides, upstream seems to answer pretty fast so I will try to push a solution there rather than a custom one in Pirouette.

Niols:
Let's believe in it; new round of tests!

Niols:
Tests successful! If you have a look at the flake checks, you'll see that every time I “break” one of the file (that is I get them out of sync), the CI complains, as expected. As soon as they are back in sync, CI is happy.

Niols:
I just opened the PR https://github.com/cachix/pre-commit-hooks.nix/pull/236 that should bring this fixed version of the `hpack` hook upstream. We will potentially iterate there on better versions but, eventually, we should settle things there. We can either wait for that to update the current PR or merge and update later on. Personally, I would go for the latter.

Niols:
Now uses upstream's hpack. I think everything is ready to merge.

Niols:
@gabrielhdt WDYT?

0xd34df00d (approved):

PR #179, opened 2023-02-08
PureSMT: update smtlib-backends to version 0.3

qaristote:
This PR updates the `PureSMT` module to use the latest version of `smtlib-backends`. The Z3 API is now calls using FFI instead of `inline-c`. Other than that, the two versions of PIrouette should be equivalent.


Niols (approved):

Niols:
Thanks!

PR #180, opened 2023-02-24
Use several SMT solvers in parallel

qaristote:
The purpose of this PR is to describe my implementation of parallelism inside Pirouette. It is not fully working, and is based on the [`niols/just-be-bourrin`](https://github.com/tweag/pirouette/compare/niols/just-be-bourrin) branch which implements checking for satisfiability at every node instead of every leaf of the symbolic execution tree. I'll comment my change in subsequent messages.

### DIFF
--- a/src/Pirouette/Symbolic/Eval.hs
+++ b/src/Pirouette/Symbolic/Eval.hs
@@ -5,7 +5,6 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
-{-# LANGUAGE InstanceSigs #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
@@ -71,7 +70,7 @@ import qualified Data.Map.Strict as M
 import Data.Maybe (mapMaybe)
 import qualified Data.Set as S
 import Data.Tuple (swap)
-import ListT.Weighted (WeightedList)
+import ListT.Weighted (WeightedList, WeightedListT)
 import qualified ListT.Weighted as ListT
 import Pirouette.Monad
 import Pirouette.Monad.Maybe
@@ -86,26 +85,26 @@ import qualified PureSMT
 
 data SymEvalSolvers lang = SymEvalSolvers
   { -- | Check whether a path is plausible
-    solvePathProblem :: CheckPathProblem lang -> Bool,
+    solvePathProblem :: CheckPathProblem lang -> IO Bool,
     -- | Check whether a certain property currently holds over a given path
-    solvePropProblem :: CheckPropertyProblem lang -> PruneResult
+    solvePropProblem :: CheckPropertyProblem lang -> IO PruneResult
   }
 
 data SymEvalEnv lang = SymEvalEnv
   { seeDefs :: PrtOrderedDefs lang,
     seeSolvers :: SymEvalSolvers lang,
-    seeAsyncSolver :: forall t res. Traversable t => t (SolverProblem lang res) -> t res,
+    seeAsyncSolver :: forall t res. Traversable t => t (SolverProblem lang res) -> IO (t res),
     seeOptions :: Options
   }
 
 -- | A 'SymEval' is equivalent to a function with type:
 --
--- > SymEvalEnv lang -> SymEvalSt lang -> [(a, SymEvalSt lang)]
+-- > SymEvalEnv lang -> SymEvalSt lang -> IO [(a, SymEvalSt lang)]
 newtype SymEval lang a = SymEval
   { symEval ::
       ReaderT
         (SymEvalEnv lang)
-        (StateT (SymEvalSt lang) WeightedList)
+        (StateT (SymEvalSt lang) (WeightedListT IO))
         a
   }
   deriving (Functor)
@@ -129,16 +128,16 @@ instance MonadFail (SymEval lang) where
   fail = error
 
 symeval ::
-  (SymEvalConstr lang, PirouetteDepOrder lang m) =>
+  (SymEvalConstr lang, PirouetteDepOrder lang m, MonadIO m) =>
   Options ->
   SymEval lang a ->
   m [Path lang a]
 symeval opts prob = do
   defs <- getPrtOrderedDefs
-  return $ runSymEval opts defs def prob
+  liftIO $ runSymEval opts defs def prob
 
 symevalAnyPathAccum ::
-  (SymEvalConstr lang, PirouetteDepOrder lang m) =>
+  (SymEvalConstr lang, PirouetteDepOrder lang m, MonadIO m) =>
   (Path lang a -> st -> st) ->
   st ->
   Options ->
@@ -147,7 +146,7 @@ symevalAnyPathAccum ::
   m (Maybe (Path lang a), st)
 symevalAnyPathAccum f s0 opts p prob = do
   defs <- getPrtOrderedDefs
-  return $ runIdentity $ ListT.firstThatAccum f s0 p $ runSymEvalWorker opts defs def prob
+  liftIO $ runIdentity . ListT.firstThatAccum f s0 p <$> runSymEvalWorker opts defs def prob
 
 -- | Running a symbolic execution will prepare the solver only once, then use a persistent session
 --  to make all the necessary queries.
@@ -157,15 +156,15 @@ runSymEval ::
   PrtOrderedDefs lang ->
   SymEvalSt lang ->
   SymEval lang a ->
-  [Path lang a]
-runSymEval opts defs st = runIdentity . ListT.toList . runSymEvalWorker opts defs st
+  IO [Path lang a]
+runSymEval opts defs st = fmap (runIdentity . ListT.toList) . runSymEvalWorker opts defs st
 
 runSymEvalRaw ::
   (SymEvalConstr lang) =>
   SymEvalEnv lang ->
   SymEvalSt lang ->
   SymEval lang a ->
-  WeightedList (a, SymEvalSt lang)
+  WeightedListT IO (a, SymEvalSt lang)
 runSymEvalRaw env st act =
   runStateT (runReaderT (symEval act) env) st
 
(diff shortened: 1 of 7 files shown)


src/Pirouette/Symbolic/Eval.hs
```diff
@@ -125,16 +128,16 @@ instance MonadFail (SymEval lang) where
...
   SymEval lang a ->
   m [Path lang a]
 symeval opts prob = do
   defs <- getPrtOrderedDefs
-  return $ runSymEval opts defs def prob
+  liftIO $ runSymEval opts defs def prob
 
 symevalAnyPathAccum ::
-  (SymEvalConstr lang, PirouetteDepOrder lang m) =>
+  (SymEvalConstr lang, PirouetteDepOrder lang m, MonadIO m) =>
```
qaristote, line 140: `runSymEvalWorker` (called in the body of this function now returns an `IO` computation so we need the resulting monadic type `m` to be an instance of `MonadIO`.

src/Pirouette/Symbolic/Eval.hs
```diff
@@ -83,25 +85,26 @@ import qualified PureSMT
...
+    solvePathProblem :: CheckPathProblem lang -> IO Bool,
     -- | Check whether a certain property currently holds over a given path
-    solvePropProblem :: CheckPropertyProblem lang -> PruneResult
+    solvePropProblem :: CheckPropertyProblem lang -> IO PruneResult
   }
 
 data SymEvalEnv lang = SymEvalEnv
   { seeDefs :: PrtOrderedDefs lang,
     seeSolvers :: SymEvalSolvers lang,
+    seeAsyncSolver :: forall t res. Traversable t => t (SolverProblem lang res) -> IO (t res),
```
qaristote, line 96: For simplicity, I just added a binding in `SymEvalEnv` for solving problems asynchronously. The bindings inside `seeSolvers` now become a bit redundant.

src/Pirouette/Symbolic/Eval.hs
```diff
@@ -83,25 +85,26 @@ import qualified PureSMT
...
 -- | A 'SymEval' is equivalent to a function with type:
 --
--- > SymEvalEnv lang -> SymEvalSt lang -> [(a, SymEvalSt lang)]
+-- > SymEvalEnv lang -> SymEvalSt lang -> IO [(a, SymEvalSt lang)]
 newtype SymEval lang a = SymEval
   { symEval ::
       ReaderT
         (SymEvalEnv lang)
-        (StateT (SymEvalSt lang) WeightedList)
+        (StateT (SymEvalSt lang) (WeightedListT IO))
```
qaristote, line 107: Most of the `IO` gets hidden inside the main monad in use in this file.

src/Pirouette/Symbolic/Eval.hs
```diff
@@ -216,25 +219,34 @@ instance (SymEvalConstr lang) => Alternative (SymEval lang) where
...
+    -- solve the problems asynchronously
+    resultsTraversable <- seeAsyncSolver env @(Compose WeightedList (Compose ((,) (SymEvalSt lang)) t)) problemsTraversable
+    let -- get the correct type back
+        results = swap . getCompose <$> getCompose resultsTraversable
+    return $ ListT.mapWeightedListT (pure . runIdentity) results
+
+learn :: forall lang. (SymEvalConstr lang) => [C.Constraint lang SymVar] -> SymEval lang ()
+learn cs = do
+  result <- fmap runIdentity $ solveAll $ Identity <$> pruneOrGetPathProblem cs
+  unless result empty
```
qaristote, line 249: We split `learn` into two sub-functions, `pruneOrGetPathProblem` and `solveAll` (described above). We get a weighted list of results, one for each problem passed to `solveAll`, corresponding to a branch whose constraints we didn't previously find to be inconsistent, and kill off the branches with a negative result (meaning the solver found the constraints to be inconsistent).

src/Pirouette/Symbolic/Eval.hs
```diff
@@ -216,25 +219,34 @@ instance (SymEvalConstr lang) => Alternative (SymEval lang) where
               (runSymEvalRaw env st xs, runSymEvalRaw env st ys)
        in xs' <|> ys'
 
--- | Learn a new constraint and add it as a conjunct to the set of constraints of
---  the current path, then checks whether it leads to an inconsistency and fails
--- if it does.
-learn :: (SymEvalConstr lang) => [C.Constraint lang SymVar] -> SymEval lang ()
-learn cs = do
+pruneOrGetPathProblem :: (SymEvalConstr lang) => [C.Constraint lang SymVar] -> SymEval lang (SolverProblem lang Bool)
```
qaristote, line 222: `pruneOrGetPathProblem` takes the current state and outputs a (weighted) list of each possible new state obtained when adding one of the constraints taken as input; if the constraint is inconsistent with the state the branch is killed off and doesn't appear in the resulting list.

src/Pirouette/Symbolic/Eval.hs
```diff
@@ -216,25 +219,34 @@ instance (SymEvalConstr lang) => Alternative (SymEval lang) where
...
+solveAll (SymEval f) = SymEval $
+  ReaderT $ \env -> StateT $ \st -> ListT.Action @IO $ do
+    problems <- ListT.runActions $ runStateT (runReaderT f env) st
+    let -- change the type to get a Traversable instance
+        problemsTraversable = Compose $ Compose . swap <$> problems
+    -- solve the problems asynchronously
+    resultsTraversable <- seeAsyncSolver env @(Compose WeightedList (Compose ((,) (SymEvalSt lang)) t)) problemsTraversable
+    let -- get the correct type back
+        results = swap . getCompose <$> getCompose resultsTraversable
+    return $ ListT.mapWeightedListT (pure . runIdentity) results
```
qaristote, line 244: Informally, `solveAll`
- takes a `WeightedList` of pairs of `t (SolverProblem lang res)` and `SymEvalSt lang`, 
- transforms it so that the overall functor wrapping the problem type `WeightedList o (,) (SymEvalSt lang) o t`) is recognized as `Traversable`, 
- solves the problems asynchronously, 
- transform the results back into a `WeightedList`
- make the `WeightedList` into a `WeightedListT IO` so we can wrap the whole computation inside the `SymEval` monad.

src/Pirouette/Symbolic/Eval.hs
```diff
@@ -311,12 +323,30 @@ symEvalOneStep t@(R.App hd args) = case hd of
...
+        -- consumeFuel
+        -- signalEvaluation
+        -- lift $ learnBranches branches
+        --
+        asum $
+          flip map branches $ \(SMT.Branch additionalInfo newTerm) -> do
+            lift $ learn additionalInfo
+            consumeFuel
+            signalEvaluation
+            pure newTerm
```
qaristote, line 337: This is how branching is handled inside Pirouette. For each branch, we add the corresponding constraints to each possible state (gathered in the `WeightedList` inside the `SymEval` monad) and prune the (current state + current branch) if the constraints are inconsistent with the state. Asking the solver to check for this consistency is done in parallel (i.e., up to one solver for each possible current state). Finally, once we've done this for every branch (sequentially), we gather all the possible new states with `asum`. In a nutshell, the scheduling can be described as
```
solve st1 cst1 || solve st1 cst2 || ...;
solve st2 cst1 || solve st2 cst2 || ...;
...
```

src/Pirouette/Symbolic/Eval.hs
```diff
@@ -311,12 +323,30 @@ symEvalOneStep t@(R.App hd args) = case hd of
...
+      --
+      -- let problemsWithTerms = asum $
+      --       flip map branches $ \(SMT.Branch additionalInfo newTerm) -> do
+      --         problem <- pruneOrGetPathProblem additionalInfo
+      --         pure $ (newTerm, problem)
+      -- (newTerm, result) <- lift $ solveAll problemsWithTerms
```
qaristote, line 343: Another way to do the same thing is described is this comment. Now we first gather the problems for all the possible pairs (current state, branch constraint), and solve all these problems asynchronously in one go.

src/Pirouette/Symbolic/Prover.hs
```diff
@@ -81,15 +81,15 @@ proveUnbounded opts = prove (opts {shouldStop = const False})
 --  means that on all paths that we looked at they were either verified or
 --  they reached the stopping condition.
 proveAny ::
-  (SymEvalConstr lang, PirouetteDepOrder lang m) =>
+  (SymEvalConstr lang, PirouetteDepOrder lang m, MonadIO m) =>
   Options ->
   (Path lang (EvaluationWitness lang) -> Bool) ->
   Problem lang ->
   m (Maybe (Path lang (EvaluationWitness lang)))
```
qaristote, line 88: Here I'm not sure what the monad is in practice. Adding the `MonadIO m` is enough to get things to compile so I only did that.

src/Pirouette/Symbolic/Prover/Runner.hs
```diff
@@ -46,9 +45,9 @@ runIncorrectnessLogic ::
   Options ->
   PrtUnorderedDefs lang ->
   IncorrectnessParams lang ->
-  IncorrectnessResult lang
-runIncorrectnessLogic opts prog parms =
-  runIdentity $ execIncorrectnessLogic (proveAny opts isCounter) prog parms
+  IO (IncorrectnessResult lang)
```
qaristote, line 48: Same comment, it seems adding the `IO` wrapper here is enough for the code to compile.

src/PureSMT.hs
```diff
@@ -58,48 +58,33 @@ instance Default Options where
...
---
--- At least one pool is started for each value of @Ctx domain@. Note that
--- multiple pools of SMT solvers might be created for the same value of
--- @Ctx domain@ if @solve@ is called from different modules, or if GHC
--- optimizations fail to apply CSE over calls to solve in the same module.
-{-# NOINLINE solveOpts #-}
-solveOpts :: forall domain res. Options -> Solve domain => Ctx domain -> Problem domain res -> res
-solveOpts opts ctx = unsafePerformIO $ do
-  -- we end up with a list of MVars, which we will protect in another MVar.
-  allProcs <- initAll @domain opts ctx >>= newMStack
```
qaristote, line 75: Since we're propagating `IO` back into Pirouette's core, there's no need to call `initAll` when calling `solveOpts`, so I split these two functions.

src/PureSMT.hs
```diff
@@ -58,48 +58,33 @@ instance Default Options where
...
-  return $ \problem -> unsafePerformIO $ do
-    ms <- popMStack allProcs
-    r <- withMVar ms $ \solver -> do
-      -- TODO: what happens in an exception? For now, we just loose a solver but we shouldn't
-      -- add it to the pool of workers and just retry the problem. In a future implementation
-      -- we could try launching it again
-      r <- solveProblem @domain problem solver
-      return r
-    pushMStack ms allProcs
+solve :: forall domain res t. Solve domain => MStack (MVar X.Solver) -> Traversable t => t (Problem domain res) -> IO (t res)
```
qaristote, line 63: The `solve` function now only takes as input the pool of worker and the problems to solve. The solver options are used when creating the pool of worker inside `initAll`.

src/PureSMT.hs
```diff
@@ -58,48 +58,33 @@ instance Default Options where
...
+      -- TODO: each init creates its own config but they could be shared
+      -- this would be especially useful if the solver options are passed
+      -- when creating the configuration instead of inside the initSolver
+      -- function
+      s <- X.initZ3Instance (debug opts)
+      initSolver @domain ctx s
+      newMVar s
   where
-    nWorkers :: Int
-    nWorkers = maybe numCapabilities ensurePos (numWorkers opts)
```
qaristote, line 101: `numCapabilties` is only evaluated at compile time and the result may differ from what's actually available at run time, hence `getNumCapabilities` is preferred.

src/PureSMT.hs
```diff
@@ -58,48 +58,33 @@ instance Default Options where
...
-  s <- X.initZ3Instance (debug opts)
-  initSolver @domain ctx s
-  newMVar s
+  pushMStack ms allProcs
+  return r
+
+initAll :: forall domain. Options -> Solve domain => Ctx domain -> IO (MStack (MVar X.Solver))
+initAll opts ctx =
+  newMStack =<< do
+    nWorkers <- ensurePos <$> maybe getNumCapabilities return (numWorkers opts)
```
qaristote, line 78: Note that `getNumCapabilities` evaluates to `1` on my laptop, so I actually had to override `numWorkers` to get multiple workers. `setNumCapabilities` might also be need to tell Haskell we're running things in parallel.

qaristote (commented):
As @Niols pointed out, `PureSMT` is not so pure anymore, so we might want to change its name ...

